PR #5, opened 2017-03-20
Feature/join population datasets

seb231:
* Load historic datasets
* Create `:state` from `:need` and `:placement`
* Create a row per individual
* Add ids for each individual
* Create a row per simulation and per individual

### DIFF
--- a/src/witan/send/schemas.clj
+++ b/src/witan/send/schemas.clj
@@ -42,10 +42,10 @@
                            [:population s/Int]]))
 
 (def SENDSchemaIndividual
-  (make-ordered-ds-schema [[:id s/Int]
-                           [:year (s/constrained s/Int u/year?)]
+  (make-ordered-ds-schema [[:year (s/constrained s/Int u/year?)]
                            [:age (s/constrained s/Int u/SENDage?)]
                            [:state s/Keyword]
+                           [:id s/Int]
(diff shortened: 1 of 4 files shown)


otfrom:
using def to hold data you are loading from files off disk is a bad idea as a problem with the file will mean that the repl won't start

Eleonore9:
Thanks @otfrom I changed that!

Eleonore9:
OK, I put `repeat` back instead of `cycle`!

src/witan/send/schemas.clj
```diff
@@ -42,10 +42,11 @@
...
 
 (def SENDSchemaIndividual
-  (make-ordered-ds-schema [[:id s/Int]
-                           [:year (s/constrained s/Int u/year?)]
+  (make-ordered-ds-schema [[:year (s/constrained s/Int u/year?)]
                            [:age (s/constrained s/Int u/SENDage?)]
-                           [:state s/Keyword]]))
+                           [:state s/Keyword]
+                           [:sim-num s/Int]
+                           [:id s/Int]]))
```
seb231: I'm sure there's a good reason for it, but why was the order changed?

Eleonore9: @seb231 because we're addind `:sim-num` and then `:id` I think they end up being the two last columns.

PR #7, opened 2017-03-28
Add extra population to historic population

seb231:
Combine the historic population with the extra population due to population growth.

### DIFF
--- a/test/witan/send/send_test.clj
+++ b/test/witan/send/send_test.clj
@@ -34,8 +34,8 @@
                :age   [0 0 0 0 0 0 2 2]
                :state [:Non-SEND :Non-SEND :Non-SEND :Non-SEND
                        :Non-SEND :Non-SEND :PSI-Mainstream :PSI-Mainstream]
-               :id    [1 1 2 2 3 3 4 4]
-               :sim-num [1 2 1 2 1 2 1 2]}))
+               :sim-num [1 2 1 2 1 2 1 2]
+               :id    [1 1 2 2 3 3 4 4]}))
 
 (def historic-data
   (ds/dataset {:year  [2016 2016 2016]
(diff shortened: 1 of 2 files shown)


seb231:
lgtm

test/witan/send/send_test.clj
```diff
@@ -73,10 +73,33 @@
...
-               :id [5 5 6 6 7 7 8 8 9 9 10 10 11 11 12 12 13 13 14 14 15 15 16 16
-                    17 17 18 18 19 19 20 20 21 21 22 22 23 23 24 24]
                :sim-num [1 2 1 2 1 2 1 2 1 2 1 2 1 2 1 2 1 2 1 2 1 2 1 2 1 2 1 2 1 2
-                         1 2 1 2 1 2 1 2 1 2]}))
+                         1 2 1 2 1 2 1 2 1 2]
+               :id [5 5 6 6 7 7 8 8 9 9 10 10 11 11 12 12 13 13 14 14 15 15 16 16
+                    17 17 18 18 19 19 20 20 21 21 22 22 23 23 24 24]}))
+
+(def population-total
+  (ds/dataset {:year [2014 2014 2014 2014 2014 2014 2016 2016
```
acron0: This scares me a lot. How come this stuff isn't being loaded from a CSV?

seb231: Possible, but would require us to make the file, then (I think) load it and run a number of other functions on it. What precisely is bad about it?

acron0: Having columns laid out like this makes it almost impossible to parse a single row, which obviously makes changes hairy and fragile. As a CSV it'd be way easier to grok. Also, in other projects we've loaded CSVs.

seb231:
After discussion with @acron0 :shipit:

PR #8, opened 2017-03-29
Data prep - Account for population change

Eleonore9:
Calculates the number of extra individuals who will be added to the population over the course of the projection due to population growth, and returns a dataset with a row for each of these individuals for each simulation.

### DIFF
--- a/src/witan/send/send.clj
+++ b/src/witan/send/send.clj
@@ -6,7 +6,8 @@
             [schema.core :as s]
             [witan.send.schemas :as sc]
             [clojure.core.matrix.dataset :as ds]
-            [witan.datasets :as wds]))
+            [witan.datasets :as wds]
+            [witan.send.utils :as u]))
 
 ;;Inputs
 (definput historic-0-25-population-1-0-0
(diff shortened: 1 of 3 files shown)


src/witan/send/send.clj
```diff
@@ -135,7 +159,61 @@
...
 
+(defn lag
+  "Dataset must be kept chronologically ordered in order to work correctly.
+   Shifts the specified column down by the specified number of rows, filling
+   in zeros at the top of the column and dropping the shifted values at the end."
+  ([dataset col-key]
+   (-> dataset
+       (ds/to-map)
+       (col-key)
+       (#(cons 0 %))
```
acron0: I can see why you've done this but it's ugly

src/witan/send/send.clj
```diff
@@ -135,7 +159,61 @@
...
+       (butlast)))
+  ([dataset col-key lag-amount]
+   (-> dataset
+       (ds/to-map)
+       (col-key)
+       (#(concat (repeat lag-amount 0) %))
+       (as-> d (drop-last lag-amount d)))))
+
```
acron0: Both the lamba and `as->` are here to work around the thread-first so I suggest a two-step approach instead

src/witan/send/send.clj
Eleonore9: @acron0 could you be more descriptive

acron0: Don't include `cons` in the thread-first

PR #9, opened 2017-03-30
Transition matrices

Eleonore9:
* In `transition-matrix-1-0-0`, get the probabilities for the transitions matrix out of a csv to create a dataset.
* In `adjust-joiners-transition-1-0-0`, depending on user params adjust probabilities inside the matrix
* In `apply-state-changes-1-0-0`, before the transition matrix is used change the format to accommodate for the use of the `markov-chains` library (we'll use later)


src/witan/send/schemas.clj
```diff
@@ -30,9 +30,51 @@
...
+             :SPLD-Mainstream
+             :SPLD-Special
+             :SPLCD-Other
+             :UO-Mainstream
+             :UO-Special
+             :UO-Other
+             :Too-old])
+
+(def SendStatesSchema
+  [(s/enum :Non-SEND :ASD-Mainstream :ASD-Special :ASD-Other
```
acron0: `(apply s/enum States)`

src/witan/send/schemas.clj
```diff
@@ -30,9 +30,51 @@
...
+
+(def ProbabilitiesSchema
+  (reduce merge
+          (map #(hash-map % double) States)))
+
+(def TransitionMatrixSchema
+  {SendStatesSchema ProbabilitiesSchema})
+
+(def AgeGroupsSchema
+  (s/enum :1 :2 :3 :4 :5 :6 :7 :8 :9 :10 :11 :12 :13 :14
```
acron0: Can we find a more programmatic way of doing this?

Eleonore9:
Afaik the structure of the transition matrix is "dictated" by what is expected by the library we want to use to implement the Markov model: https://github.com/rm-hull/markov-chains.
Let's talk about it on Monday @acron0 !

src/witan/send/schemas.clj
Eleonore9: :+1:

src/witan/send/schemas.clj
Eleonore9: What about `(apply s/enum (map #((comp keyword str) %) (range 1 27)))`?

acron0:
@eleonore I'd still prefer if that massive EDN file didn't exist - I can't see where it's being used?

Eleonore9:
@acron0 I'm using it in the tests now!
We need this data as it's an input for the model. What shape should it be in?

src/witan/send/utils.clj
```diff
@@ -19,3 +19,45 @@
...
                                (sort-by (apply juxt col-key))
                                ds/dataset)))
+
+(defn probs-1 [dataset]
+  (reduce merge (map (fn [{:keys [state probability]}] {state probability})
+                     (-> dataset
+                         (ds/select-columns [:state :probability])
+                         ds/row-maps))))
+
+(defn trans-mat [probabilities]
```
acron0: Could this function 'build' the map by reducing over the States sequence in schema? 
https://github.com/MastodonC/witan.send/pull/9/files#diff-eb5f87c92704c5d4a503be087cc2aee1R33

src/witan/send/utils.clj
```diff
@@ -19,3 +19,45 @@
...
+   [:SPLD-Mainstream] probabilities
+   [:SPLD-Special] probabilities
+   [:SPLCD-Other] probabilities
+   [:UO-Mainstream] probabilities
+   [:UO-Special] probabilities
+   [:UO-Other] probabilities
+   [:Too-old] probabilities})
+
+
+(defn full-trans-mat [dataset]
```
acron0: Feels like this data structure could also be generated/built programmatically.

src/witan/send/model.clj
```diff
@@ -23,7 +22,7 @@
 
    ;;Loop
    [:select-starting-population :apply-state-changes]
-   [:get-transition-matrix :apply-state-changes]
+   [:adjust-joiners-transition :apply-state-changes]
```
seb231: So these changes haven't yet been applied to the flowchart? Keen to close the branch sooner rather than later, so if that's the case, let's just make that a new story/branch

src/witan/send/utils.clj
```diff
@@ -19,3 +19,26 @@
...
+  (let [age-rows (->> (ds/column dataset :age)
+                      (filter-indexed #(when (= %2 age) %1))
+                      (ds/select-rows dataset)
+                      (ds/row-maps))]
+    (reduce (fn [acc field]
+              (let [field-rows (filter #(= (:from-state %) field) age-rows)]
+                (assoc acc [field]
+                       (reduce (fn [acc row]
+                                 (assoc acc (:to-state row) (:probability row)))
+                               {} field-rows)))) {} fields)))
```
mantree: select-keys [:to-state :probability] perhaps?

src/witan/send/model.clj
Eleonore9: @seb231 Yes, let's create a separate story for that!

PR #12, opened 2017-04-07
Join new dataset out of the loop back with the global population

Eleonore9:
Last step of the loop is to join back the new population (that underwent transition changes) back to the total population.

+ Test added
+ Input data reduced in the test to bypass memory issue in circleCI


Eleonore9:
Looking good! :sunglasses:

PR #13, opened 2017-04-10
Group send projections

Eleonore9:
Group by age/placement/need and calculate average and confidence interval over simulations

### DIFF
--- a/src/witan/send/send.clj
+++ b/src/witan/send/send.clj
@@ -7,6 +7,7 @@
             [witan.send.schemas :as sc]
             [clojure.core.matrix.dataset :as ds]
             [witan.datasets :as wds]
+            [witan.datasets.stats :as wst]
             [witan.send.utils :as u]
             [markov-chains.core :as mc]))
 
@@ -320,7 +321,8 @@
      :current-year-in-loop current-year-in-loop}))
 
 (defworkflowfn append-to-total-population-1-0-0
-  "Adds the newly calculated states of the individuals for this year to the total population dataset"
+  "Adds the newly calculated states of the individuals for
+   this year to the total population dataset"
   {:witan/name :send/append-to-total-population
    :witan/version "1.0.0"
    :witan/input-schema {:total-population sc/SENDSchemaIndividual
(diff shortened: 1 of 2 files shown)


src/witan/send/send.clj
```diff
@@ -342,12 +345,61 @@
...
+
+(defn confidence-interval-and-mean
+  "Takes in a dataset of send projections and outputs average and confidence interval boundaries."
+  [grouped-projection]
+  (->> grouped-projection
+       (pmap (fn [[k v]] (let [state (:state k)
+                               sims-counts (frequencies (ds/column v :sim-num))
+                               mean (/ (apply + (vals sims-counts)) (count sims-counts))
+                               sorted-counts (sort (vals sims-counts))
+                               low-ci (quantile* 0.25 sorted-counts)
```
Eleonore9: Thanks @sunnytownsend !

src/witan/send/send.clj
```diff
@@ -342,12 +345,61 @@
...
+  "Takes in a dataset of send projections and outputs average and confidence interval boundaries."
+  [grouped-projection]
+  (->> grouped-projection
+       (pmap (fn [[k v]] (let [state (:state k)
+                               sims-counts (frequencies (ds/column v :sim-num))
+                               mean (/ (apply + (vals sims-counts)) (count sims-counts))
+                               sorted-counts (sort (vals sims-counts))
+                               low-ci (quantile* 0.025 sorted-counts)
+                               high-ci (quantile* 0.975 sorted-counts)
+                               split-keyword (fn [kw] (clojure.string/split (name kw) #"-"))
```
seb231: I appreciate that it may be easier to do this separation of need and placement here, but is better if this was split out into a different function?

acron0: I encouraged this for the purposes of speed. In fact, I think we might see a few more corners being cut as we trade principles for efficiency!

seb231: Speed in terms of coding up or run-time? I'm just concerned this isn't represented in the name of the function or in doc-string

acron0: In terms of run-time. But you're correct, this should be reflected.

seb231: OK as long as it's reflected somehow I'm happy with this then

Eleonore9:
@seb231 afaik we give the user an average number of children per send need/placement. And include the CI as extra information on how the actual numbers are most likely to vary from the mean.

seb231:
:shipit:

PR #15, opened 2017-04-13
Add a model description with updated fowchart

Eleonore9:


acron0:
@Eleonore9 We had a brief conversation about how the joiners step should be outside the loop - is this something you intend to adjust?

seb231:
Does 'multiplier' need explaining? Multiplier to what?

Eleonore9:
Good point @seb231!

acron0:
I agree with seb's comment - some example numbers and what they mean (e.g. "1.0" is 100% of the defaault value, therefore "0.2" is an...80% reduction? IANAM!)

Eleonore9:
Good idea @acron0 ! Updated the text again.

PR #20, opened 2018-01-09
Feature/output to chart

seb231:
Removed redundant data files, updated valid-settings with groupings for Sankey plotting and removed from code, output ggplot Sankey plot to file for each year, created model parameter to stop acceptance test needlessly outputting files


src/clj/witan/send/charts.clj
```diff
@@ -30,57 +31,6 @@
...
-            :PRU "Other"
-            :OOE "Other"
-            :FEC "Further education"
-            :ISC "Special"
-            :ISCR "Special"
-            :IT "Other"
-            :NON_SEND "Other"))
-
 (defn sankey [{:keys [title] :or {title ""}} df]
   (gg4clj/render
```
elisehuard: interesting use of gg4clj - wasn't expecting it out of a gorilla form.
maybe we should extract the non-gorilla bits (ie running the r code and outputting to a file) into a util

src/clj/witan/send/send.clj
```diff
@@ -334,80 +341,93 @@
...
+        (let [columns [:calendar-year :ay-group :mean :std-dev :iqr :min :low-ci :q1 :median :q3 :high-ci :max]]
+          (->> (mapcat (fn [output year]
+                         (map (fn [[ay-group stats]]
+                                (-> (medley/map-vals round stats)
+                                    (assoc :ay-group ay-group :calendar-year year)))
+                              (:total-in-send-by-ay-group output))) send-output (range initial-projection-year 3000))
+               (map (apply juxt columns))
+               (concat [(map name columns)])
+               (csv/write-csv writer))))
+      (run! #(ch/sankey-transitions transitions-data % valid-settings) years)))
```
elisehuard: just to be sure I understand, this line and lines around 360 are the only changes in this block?

src/clj/witan/send/charts.clj
seb231: I thought about this, but at least at the moment this Sankey function is so specific to its use here. In the future we could maybe recreate it with more args to add more flexibility

src/clj/witan/send/send.clj
seb231: Yes all of the the code output and the new plot file output have been put in an `if` so when running tests it's not creating additional files

src/clj/witan/send/charts.clj
elisehuard: fair

PR #25, opened 2018-01-29
Pop line chart

mikeyford:


src/R/send-charts.R
```diff
@@ -0,0 +1,203 @@
+library(dplyr)
+library(ggplot2)
+library(reshape2)
+library(stringr)
+
+### Variables for all charts ###
+
+df_historical <- read.csv("../../target/historic-data.csv")
```
seb231: I suspect that this "probably" would work from someone else's machine, but I would prefer if we could add absolute paths. Perhaps an RProject is the solution?

src/R/send-charts.R
```diff
@@ -0,0 +1,203 @@
...
+df_historical_ay <- df_historical %>%
+  group_by(academic.year.1, calendar.year) %>%
+  count()
+
+names(df_historical_ay)[1] <- 'academic.year'
+names(df_historical_ay)[3] <- 'mean'
+
+df_ay <- rbind(df_historical_ay[,c(2,1,3)], df_projected_ay)
+
+df_ay <- transform(df_ay, academic.year = as.numeric(academic.year))
```
seb231: Could these bits be added to the code below in the dply section, thereby threading the whole bit?

src/clj/witan/send/charts.clj
```diff
@@ -107,4 +107,74 @@
...
+  (move-file "Rplots.pdf" (str "target/" title "_Probability.pdf")))
+
+(defn get-total-historical-send-pop [year transitions-data]
+  (count (map :academic-year-1 (filter #(= (:calendar-year %) year) (filter #(not= (:setting-1 %) :NONSEND) transitions-data)))))
+
+(defn is-send-and-is-year? [row year]
+  (and (not= (:setting-1 row) :NONSEND)
+        (= (:calendar-year row) year)))
+
+(defn foo [year transitions-data]
```
seb231: Perhaps a better name? :)

src/clj/witan/send/charts.clj
```diff
@@ -107,4 +107,74 @@
...
+                         (inc (last historic-years))))
+        historic-pops (map #(get-total-historical-send-pop % historic-data) historic-years)
+        years (into [] (concat historic-years projected-years))
+        means (into [] (concat historic-pops (map int (map :mean future-projections-data))))
+        low-cis (into [] (concat (repeat (count historic-years) :NA) (map :low-ci future-projections-data)))
+        high-cis (into [] (concat (repeat (count historic-years) :NA) (map :high-ci future-projections-data)))
+        gg4clj-data-frame {:year years :mean means :min low-cis :max high-cis}
+        min-year (first historic-years)
+        max-year (last projected-years)
+        max-max (apply max (map :max future-projections-data))]
```
seb231: Is this max population?

src/clj/witan/send/charts.clj
```diff
@@ -107,4 +107,74 @@
...
+        low-ci (vec  (map #(/ % 1000000) (map :low-ci cost-data)))
+        q1 (vec (map #(/ % 1000000) (map :q1 cost-data)))
+        median  (vec (map #(/ % 1000000) (map :median cost-data)))
+        q3 (vec (map #(/ % 1000000) (map :q3 cost-data)))
+        high-ci (vec (map #(/ % 1000000) (map :high-ci cost-data)))
+        gg2clj-data-frame {:year proj-years :ymin low-ci :lower q1 :middle median :upper q3 :ymax high-ci}
+        max-max (last high-ci)
+        min-year (first proj-years)
+        max-year (last proj-years)
+        ]
```
seb231: tidy

src/clj/witan/send/charts.clj
```diff
@@ -107,4 +107,74 @@
...
+    (move-file "Rplots.pdf" "target/Total_Population.pdf"))
+ 
+(defn send-cost-plot
+  [cost-data years]
+  (let [proj-years (vec (#(range % (+ (count cost-data) %)) (inc (last years))))
+        low-ci (vec  (map #(/ % 1000000) (map :low-ci cost-data)))
+        q1 (vec (map #(/ % 1000000) (map :q1 cost-data)))
+        median  (vec (map #(/ % 1000000) (map :median cost-data)))
+        q3 (vec (map #(/ % 1000000) (map :q3 cost-data)))
+        high-ci (vec (map #(/ % 1000000) (map :high-ci cost-data)))
```
seb231: Sorry I haven't looked elsewhere, but why `(/ % 1000000)`?

src/clj/witan/send/charts.clj
mikeyford: ok

src/clj/witan/send/charts.clj
mikeyford: This gives the costs in millions, as with the Excel chart it copies,  and avoids an enormous y-axis.

src/clj/witan/send/charts.clj
mikeyford: it sets the upper limit on the y-axis by taking the largest max value. I'll make this more explicit.

src/clj/witan/send/charts.clj
mikeyford: Ooops That's one demo function you wrote. I need to take that out.

src/R/send-charts.R
mikeyford: I don't think so. That's changing the type of a column in the dataframe so it's not a dplyr function.

src/R/send-charts.R
mikeyford: I'm fairly confident this is the cleanest way to work because this Rscript is explicitly called from the location /src/R/ by Clojure which sets the workspace for the script there. I considered a few ways of doing this and this seemed like the safest and simplest option, while keeping this production code in a src directory.

src/R/send-charts.R
```diff
@@ -0,0 +1,203 @@
...
+
+plot_ay <- melt(plot_ay, "calendar.year")
+y_max <- max(plot_ay$value)
+
+ggplot(plot_ay, aes(x=calendar.year, y=value, group=variable)) +
+  geom_line(aes(color=variable)) +
+  geom_point(aes(color=variable)) + 
+  scale_x_discrete(name='Calendar Year') +
+  scale_y_continuous(name='SEND Population') + 
+  scale_color_manual(name = "NCY", values = c("cyan", "coral", "dodgerblue", "hotpink", "green4"), labels = c("0 and under", "1 to 6", "7 to 11", "12 to 13", "14 and up")) +
```
sunnytownsend: split this across multiple lines to make more readable

src/R/send-charts.R
```diff
@@ -0,0 +1,203 @@
...
+### Projected count by need type ###
+
+df_projected_need <- read.csv("../../target/output-need.csv")
+df_projected_need <- df_projected_need[,c("calendar.year","need","mean")]
+
+df_historical_need <- df_historical %>%
+  group_by(need.1, calendar.year) %>%
+  count()
+
+names(df_projected_need)[2] <- 'Need'
```
sunnytownsend: as above for the use of numbers rather than column names on line 118 and below

src/R/send-charts.R
```diff
@@ -0,0 +1,203 @@
...
+### Projected special setting count ###
+   
+df_projected_set <- read.csv("../../target/output-setting.csv")
+df_projected_set <- df_projected_set[,c("calendar.year","setting","mean")]
+
+df_historical_set <- df_historical %>%
+  group_by(setting.1, calendar.year) %>%
+  count()
+
+names(df_projected_set)[2] <- 'Setting'
```
sunnytownsend: ditto regarding column numbers

src/R/send-charts.R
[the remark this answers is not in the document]
mikeyford: It's because `academic.year` is a character type when it's loaded from the csv....was causing me much pain trying to implement lines 38 to 42 until I realised this.

src/R/send-charts.R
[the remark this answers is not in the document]
seb231: It does, I've been looking at this and actually because this is being run with RScript I wonder if RProject file may be redundant here? Anyone Rscript takes the path of the script as the root, so using `../../` is always going to work provided everything else in the model does

PR #26, opened 2018-01-31
Fix/all valid transitions retained

seb231:
Previously states were being checked by aging on a year and then looking for the mover/leaver/joiner rates once the transition occured. So a state was stored as [to-year from-state]. This meant that if a state was not valid for a certain year there would be no rates to check. To resolve this states are now stored as [from-year from-state] and a validation step has been added to warn of this occurring in the future.


src/clj/witan/send/send.clj
```diff
@@ -433,110 +439,126 @@
...
+                                          (states/state (:need-1 %) (:setting-1 %))
+                                          (states/state (:need-2 %) (:setting-2 %))))
+                                   distinct)
+        transform-projection (->> projection
+                                  keys
+                                  (map #(vec (drop 1 %)))
+                                  distinct)]
+    (if (every? (fn [transition] (transition-present? transition transform-projection)) transform-transitions)
+      (println "Not every historic transition present in projection. Consider checking valid state input"))
+    (if (= output true)
```
elisehuard: `(if output`

src/clj/witan/send/send.clj
```diff
@@ -433,110 +439,126 @@
...
+      (let [valid-settings (assoc (->> (ds/row-maps valid-setting-academic-years)
+                                       (reduce #(assoc %1 (:setting %2) (:setting->group %2)) {}))
```
elisehuard: down the line (not too long down the line) it might be good to break this up into little functions, with the appropriate let processing close to the outputs they belong to.  would also avoid the massive diff :)
possible even going further: having a function for all the output production, so that the `if` is readable
```
(if output
  (produce-all-the-output x y z)
  send-output)
```

seb231: good call!

PR #27, opened 2018-02-02
Feature/apply modified joiners elsewhere

seb231:
Take the removed joiners from specific states and apply them elsewhere so the same overall joiner rate is maintained. Where to redistribute these joiners is now provided as an additional optional input file.

Run-model is now multi-arity able to take the "default" two arguments or additional two more, allowing a modification the joiner rate of settings specified in a new input file

### DIFF
--- a/src/clj/witan/send/send.clj
+++ b/src/clj/witan/send/send.clj
@@ -391,7 +391,6 @@
                    (do (println "Reducing...")
                        (transduce (map #(map :model %)) (reduce-rf iterations valid-states setting-cost-lookup) projection))))
         projection (apply concat projections)]
-    (output-transitions "target/transitions.edn" projection)
     ;;    (println mover-beta-params)
     (println "Combining...")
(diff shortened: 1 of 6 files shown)


src/clj/witan/send/send.clj
```diff
@@ -178,46 +184,36 @@
...
-  (vector cy ay :NONSEND state))
-
-(defn modify-transition-count [transitions k modifier v]
-  (if (contains? transitions k)
-    (update transitions k #(u/int-ceil (modifier % v)))
+(defn generate-transition-key [cy ay need setting]
+  (vector cy ay :NONSEND (states/state need setting)))
+
+(defn update-ifelse-assoc [m k arithmetic-fn v]
+  (if (contains? m k)
```
elisehuard: possibly this could be replaced by 
```
(assoc m k (fnil #(arithmetic-fn % v) v) (get m k))
```

PR #28, opened 2018-02-07
Feature/apply two transition rates

seb231:


src/clj/witan/send/send.clj
```diff
@@ -234,26 +293,25 @@
...
+                                           result (reduce (fn [m k] (modify-transitions m k * modify-transition-by)) convert states-to-change)]
+                                       (mapcat (fn [[k v]] (u/back-to-transitions-matrix k v)) result)))
+        _ (if (nil? modified-transition-matrix)
+            (prn "Using input transitions matrix")
+            (prn "Using modified transitions matrix"))
+        transitions (if (nil? modified-transition-matrix)
+                      (u/transitions-map transition-matrix)
+                      (u/transitions-map modified-transition-matrix))
+        transition-matrix-filtered (if (nil? modified-transition-matrix)
+                                     (filter #(= (:calendar-year %) 2016) transition-matrix)
```
elisehuard: would make 2016 a constant with an expressive name (def), likely to change in the future

seb231: 😄 Please see my next PR

src/clj/witan/send/send.clj
```diff
@@ -102,12 +102,23 @@
...
                       leaver-beta-params
                       mover-beta-params mover-state-alphas
-                      valid-year-settings] :as params}
+                      valid-year-settings] :as standard-projection}
+   {:keys [modified-joiner-beta-params modified-joiner-state-alphas
+           modified-leaver-beta-params modified-mover-beta-params
+           modified-mover-state-alphas] :as scenario-projection}
    {:keys [model transitions]} [calendar-year projected-population]]
-  (let [cohorts (step/age-population projected-population model)
+  (let [params (if (= 2000 modify-transitions-from)
```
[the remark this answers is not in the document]
seb231: Input parameters to the model cannot be optional so it needed to be something, 2000 is just the default, as it is essentially an impossible number that a user would want to I could change project from. It could just as easily be 1900, if that make it more explicit?

src/clj/witan/send/send.clj
seb231: The next PR makes the `2016` an input parameter to the model

src/clj/witan/send/send.clj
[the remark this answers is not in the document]
elisehuard: possibly, but I'd def that as well so that people don't get confused default-start-year or something

PR #30, opened 2018-02-14
Feature/make splice by generic

seb231:


src/clj/witan/send/send.clj
```diff
@@ -309,10 +323,11 @@
...
                       (u/transitions-map transition-matrix)
                       (u/transitions-map modified-transition-matrix))
-        transition-matrix-filtered (if (nil? modified-transition-matrix)
-                                     (filter #(= (:calendar-year %) 2016) transition-matrix)
-                                     (filter #(= (:calendar-year %) 2016) modified-transition-matrix))
-
+        transition-matrix-filtered (if (nil? filter-transitions-from)
+                                     nil
+                                     (if (nil? modified-transition-matrix)
+                                       (filter #(= (:calendar-year %) filter-transitions-from) transition-matrix)
```
mikeyford: Is `filter-transitions-from` a single year, or a list of years? If it's just a single year, won't this only give only one year of data in `transitions-matrix-filtered`?

seb231: yes you're right! Previously we've only wanted to filter from final year so it's only ever been one year, but it would make more sense to allow multiple years. good catch!

src/clj/witan/send/schemas.clj
```diff
@@ -24,6 +24,9 @@
 (def CalendarYear
   (s/constrained s/Int #(<= 1900 % 2100)))
 
+(def CalendarYearRange
```
mantree: Don't look right? can this just be removed?

src/clj/witan/send/send.clj
```diff
@@ -216,37 +216,49 @@
...
+  (let [start-map {:population-by-age-state initial-state
+                   :valid-setting-academic-years valid-setting-academic-years
+                   :transition-matrix original-transitions
+                   :population population
+                   :setting-cost-lookup (->> (ds/row-maps setting-cost)
+                                             (map (juxt (juxt :need :setting) :cost))
+                                             (into {}))
+                   :projected-population (->> (ds/row-maps population)
+                                              (group-by :calendar-year)
+                                              (map (fn [[k v]] [k (u/total-by-academic-year v)]))
```
mantree: Consider medley.core/map-vals

src/clj/witan/send/send.clj
```diff
@@ -216,37 +216,49 @@
...
+                                              (into {}))}]
+    (if (nil? transition-matrix-filtered)
```
mantree: consider removing the nil? and switching the branches round

src/clj/witan/send/send.clj
```diff
@@ -309,9 +323,11 @@
         transitions (if (nil? modified-transition-matrix)
                       (u/transitions-map transition-matrix)
                       (u/transitions-map modified-transition-matrix))
-        transition-matrix-filtered (if (nil? modified-transition-matrix)
-                                     (filter #(= (:calendar-year %) 2016) transition-matrix)
-                                     (filter #(= (:calendar-year %) 2016) modified-transition-matrix))
+        transition-matrix-filtered (if (nil? filter-transitions-from)
```
mantree: use when

src/clj/witan/send/send.clj
```diff
@@ -309,9 +323,11 @@
...
+                                     nil
+                                     (if (nil? modified-transition-matrix)
+                                       (mapcat (fn [year] (filter #(= (:calendar-year %) year) transition-matrix)) filter-transitions-from)
```
mantree: remove if and just use (or modified-transition-matrix transition-matrix) in the expression

src/clj/witan/send/send.clj
```diff
@@ -322,16 +338,17 @@
...
     (s/validate (sc/TransitionsMap+ valid-needs valid-settings) transitions)
     (s/validate (sc/NeedSettingCost+ valid-needs valid-settings) setting-cost)
     {:standard-projection (prep-inputs initial-state splice-ncy valid-states transition-matrix transition-matrix-filtered
-                                       population valid-setting-academic-years original-transitions setting-cost)
+                                       population valid-setting-academic-years original-transitions setting-cost
+                                       filter-transitions-from)
      :scenario-projection (if ((complement nil?) modified-transition-matrix)
-                            (prep-inputs initial-state splice-ncy valid-states modified-transition-matrix transition-matrix-filtered
-                                         population valid-setting-academic-years original-transitions setting-cost)
+                            (prep-inputs initial-state splice-ncy valid-states modified-transition-matrix
```
mantree: this is a crazy number of arguments, consider using a map for them

src/clj/witan/send/send.clj
seb231: not for just one call

src/clj/witan/send/send.clj
seb231: I feel like a map will add more code but not make anything clearer really. Lots of args here is a bit of a legacy thing, it could do with a whole makeover!

mantree: ok

src/clj/witan/send/send.clj
```diff
@@ -322,16 +337,17 @@
...
-        splice-ncy 11]
+                                 (states/calculate-valid-year-settings-from-setting-academic-years))]
     (s/validate (sc/SENDPopulation+ valid-settings) initial-send-population)
     (s/validate (sc/TransitionsMap+ valid-needs valid-settings) transitions)
     (s/validate (sc/NeedSettingCost+ valid-needs valid-settings) setting-cost)
     {:standard-projection (prep-inputs initial-state splice-ncy valid-states transition-matrix transition-matrix-filtered
-                                       population valid-setting-academic-years original-transitions setting-cost)
+                                       population valid-setting-academic-years original-transitions setting-cost
+                                       filter-transitions-from)
      :scenario-projection (if ((complement nil?) modified-transition-matrix)
```
mantree: (complement nil?) doesn't make much sense. Just say (if modified-transition-matrix

src/clj/witan/send/send.clj
```diff
@@ -309,9 +323,10 @@
         transitions (if (nil? modified-transition-matrix)
                       (u/transitions-map transition-matrix)
                       (u/transitions-map modified-transition-matrix))
-        transition-matrix-filtered (if (nil? modified-transition-matrix)
-                                     (filter #(= (:calendar-year %) 2016) transition-matrix)
-                                     (filter #(= (:calendar-year %) 2016) modified-transition-matrix))
+        transition-matrix-filtered (when filter-transitions-from
+                                     (if modified-transition-matrix
```
mantree: you can push this 'if' into an 'or' in side the expression

src/clj/witan/send/send.clj
seb231: tested, wouldn't work in this scenario, as I'm changing the shape of the data

PR #31, opened 2018-02-19
Parameter log

mikeyford:


project.clj
```diff
@@ -3,8 +3,7 @@
   :url "http://example.com/FIXME"
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :dependencies [[kixi/gg4clj "0.1.1-SNAPSHOT" :exclusion [org.clojure/clojure]]
-                 [org.clojure/clojure "1.8.0"]
+  :dependencies [[org.clojure/clojure "1.8.0"]
```
seb231: What did this do?

src/clj/witan/send/send.clj
```diff
@@ -335,6 +330,14 @@
 
         valid-year-settings (->> (ds/row-maps valid-setting-academic-years)
                                  (states/calculate-valid-year-settings-from-setting-academic-years))]
+
+    (when (boolean (resolve 'u/log))
+      (when (not= 1 modify-transition-by)
+        (u/log-info (str "Modified transitions by " modify-transition-by)))
+      (if (nil? modified-transition-matrix)
```
seb231: This is a new thing to me, but just `(if modified-transition-matrix then else)` should work here, just swap round the outputs

src/clj/witan/send/send.clj
```diff
@@ -539,6 +542,9 @@
             mover-rates-CI (map #(confidence-interval mover-rates %) years)
             ;;n-colours (vec (repeatedly (count years) ch/random-colour)) ;; alternative random colour selection
             n-colours (take (count years) ch/palette)]
+        (u/log-info (str "First year of input data: " (first years)))
+        (u/log-info (str "Final year of input data: " (inc (last years))))
+        (u/log-info (str "Final year of projection: " (+ (last years) (count (map :total-in-send send-output)))))
```
seb231: Great! 👍

src/clj/witan/send/send.clj
```diff
@@ -622,5 +628,8 @@
         (ch/ribbon-plot leaver-rates-CI "Leaver" years n-colours)
         (ch/ribbon-plot mover-rates-CI "Mover" years n-colours)
         (ch/population-line-plot transitions-data (map :total-in-send send-output))
-        (ch/send-cost-plot (map :total-cost send-output) years))))
+        (ch/send-cost-plot (map :total-cost send-output) years)
+        (u/write-log)
```
seb231: I think if `write-log` is here, it will only output if `= output? true` in the run-model fn?

src/clj/witan/send/utils.clj
```diff
@@ -352,3 +356,21 @@
...
   (int (Math/ceil n)))
+
+(defn reset-log []
+  (def log (atom []))
+  (sh/sh "git" "fetch" "--tags")
+  (swap! log conj (str "SEND Model " (:out (sh/sh "git" "describe" "--abbrev=0" "--tags"))))
+  (swap! log conj (str "Run on " (f/unparse-local (f/formatter "YYYY-MM-dd") (t/today)) " at " (f/unparse (f/formatter "HH:mm") (t/now))))
+  (swap! log conj (str (System/getProperty "os.name") " " (System/getProperty "os.version")))
+  (swap! log conj (str "Clojure " (clojure-version) ", JVM " (System/getProperty "java.vm.version")))
+  (swap! log conj (str "Using branch: " (:out (sh/sh "git" "rev-parse" "--symbolic-full-name" "--abbrev-ref" "HEAD")) "\n")))
```
seb231: lgtm, but you may want a 2nd review here?

project.clj
mikeyford: I just noticed the dependency was duplicated, so I took one of the instances out as housekeeping.

src/clj/witan/send/send.clj
acron0: @mikeyford Why do you need `(when (boolean (resolve 'u/log)) ...` ?

src/clj/witan/send/send.clj
mikeyford: Ah, yeah, good point. Do we want that, or should we output the summary in both cases? I can see arguments both ways.

src/clj/witan/send/send.clj
mikeyford: @acron0 This was because the tests on the `defworkflow` this is in were failing because u/log didn't exist when it was tested independently.

acron0: @mikeyford This suggests an issue with your dependencies. You shouldn't have to call `resolve` in code like this.

project.clj
seb231: ah ok sorry, wasn't clear until I looked at the whole file

src/clj/witan/send/send.clj
seb231: Yeah, good call... I think as we're actually returning less to screen, I think we'd *always* want the log taken. It's just stored in target isn't it? So if it wasn't important it'll just be written over next time and I am assuming it doesn't really slow anything down when this is output?

src/clj/witan/send/utils.clj
```diff
@@ -352,3 +356,21 @@
 
 (defn int-ceil [n]
   (int (Math/ceil n)))
+
+(defn reset-log []
+  (def log (atom []))
```
otfrom: you need to define log *outside* of the defn. You should never have a `def` inside a defn.

otfrom: I'd actually move the logging stuff to a witan.send.report ns and rename `log-info` to `append`

PR #35, opened 2018-02-20
Sort high-ci to get actual highest value

seb231:


src/clj/witan/send/charts.clj
```diff
@@ -183,7 +183,7 @@
         q3 (vec (map to-millions (map :q3 cost-data)))
         high-ci (vec (map to-millions (map :high-ci cost-data)))
         gg2clj-data-frame {:year proj-years :ymin low-ci :lower q1 :middle median :upper q3 :ymax high-ci}
-        highest-val (last high-ci)
+        highest-val (max high-ci)
```
otfrom: if high-ci is a seq like thing this won't give you the answer you want

PR #44, opened 2018-03-22
Feature/validate transitions

seb231:
Model now checks to see setting to setting transitions are valid


src/cljc/witan/send/params.cljc
```diff
@@ -340,10 +342,15 @@
...
+                    valid-trans (get valid-transitions setting)
                     obs (get-in observations [ay state])
                     ay-obs (get observations-per-ay ay)
-                    settings (get valid-settings [ay need])
-                    prior (->> (zipmap (vec settings) (repeat (/ 1.0 (count settings))))
+                    settings (->> (get valid-settings [ay need])
+                                  vec
+                                  (into valid-trans)
+                                  u/duplicates
+                                  vec)
```
otfrom: what is going on here? I'm not sure why there are 2 calls to `vec` and what `duplicates` is supposed to be doing.

src/cljc/witan/send/params.cljc
```diff
@@ -300,16 +300,18 @@
                                {} observations)
         prior-per-year (continue-for-latter-ays prior-per-year academic-years)]
     (reduce (fn [coll [ay state]]
-              (if-let [beta-params (merge-with +
-                                               (get-in observations [ay state])
-                                               (get prior-per-year ay))]
-                (assoc coll [ay state] beta-params)
+              (if (< 1 (count (get valid-transitions (second (s/need-setting state)))))
```
otfrom: Can we turn this into a predicate that says what it means?

src/cljc/witan/send/params.cljc
seb231: Getting the valid-settings and valid-trans into the same format then combining them, and removing the duplicates, then putting it back into the format the next part of the process expects (a vector)

otfrom: you can remove actual duplicates by doing `(into #{} <your vector>)`

src/cljc/witan/send/params.cljc
```diff
@@ -300,16 +303,18 @@
                                {} observations)
         prior-per-year (continue-for-latter-ays prior-per-year academic-years)]
     (reduce (fn [coll [ay state]]
-              (if-let [beta-params (merge-with +
-                                               (get-in observations [ay state])
-                                               (get prior-per-year ay))]
-                (assoc coll [ay state] beta-params)
+              (if (any-valid-transitions? (count (get valid-transitions (second (s/need-setting state)))))
```
otfrom: I was thinking the call here would look like `(if (any-valid-transition? valid-transitions state)`

src/cljc/witan/send/params.cljc
seb231: Whoops not removing duplicates, keeping them!

PR #50, opened 2018-05-10
Feature/validate

mikeyford:
This implements the forward chaining validation workflow for send (outlined here: https://witanblog.wordpress.com/2018/01/09/roadmap-for-implementing-send-validation/#comment-1284)

 - Each year of input data is split into separate "folds" which contain the data up to and including that year. 
- The model is run on each individual fold of input data and the results are moved into a temp dir
- The "ground truth" is added to the model predictions, and then the predictions are combined into a single csv.

### DIFF
--- a/src/clj/witan/send/validate_model.clj
+++ b/src/clj/witan/send/validate_model.clj
@@ -1,0 +1,76 @@
+(ns witan.send.validate-model
+  (:require [witan.send.test-utils :as tu]
+            [clojure.data.csv :as csv]
+            [clojure.java.io :as io]))
+
+
+
+;; get files to validate on
+
+
+;;; write validations dir
+
+(defn csv-data->maps [csv-data]
+  (map zipmap
+       (->> (first csv-data)
+            (map keyword)
+            repeat)
+       (rest csv-data)))
+
+(defn write-csv [transitions path]
+  (with-open [writer (io/writer (io/file path))]
+    (let [columns (into [] (keys (first transitions)))
+          headers (mapv name columns)
+          rows (mapv #(mapv % columns) transitions)]
+    (csv/write-csv writer (into [headers] rows)))))
+
+(defn load-transitions [path]
+  (csv-data->maps
+    (with-open [reader (io/reader (str "data/" path "transitions.csv"))]
+      (doall
+        (csv/read-csv reader)))))
+
+(defn get-unique-years [transitions]
+  (->> (map :calendar-year transitions)
+       (distinct)
+       (map #(Integer/parseInt %))))
+
+
+(defn return-up-to-fold [year-limit transitions]
+  (filter #(<= (Integer/parseInt (:calendar-year %)) year-limit) transitions))
+
+(defn return-after-fold [year-limit transitions]
+  (filter #(> (Integer/parseInt (:calendar-year %)) year-limit) transitions))
+
+
+
+
+
+(defn split-and-test [year transitions]
+  (let [train (return-up-to-fold year transitions)
+        test (return-after-fold year transitions)]
+    (write-csv test (str "validate/test" year ".csv"))))
+
+
+
+;;for each year
+  ;; train = split up to fold
+  ;; test = split after fold
+  ;; write test to csv
+  ;; run model with csvs
+  ;; store results
+
+(defn split-into-folds [transitions]
+  (let [years (get-unique-years transitions)]))
+
+
+
+
+
+
+;; split files into folds
+
+;; run model with individual folds
+  ;; shuttle results into unqiue folders
+
+;; colate results
--- a/src/clj/witan/send/validate_model.clj
+++ b/src/clj/witan/send/validate_model.clj
@@ -32,6 +32,12 @@
       (doall
         (csv/read-csv reader)))))
 
+(defn load-csv-as-maps [file]
+  (csv-data->maps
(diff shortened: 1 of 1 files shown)


src/clj/witan/send/validate_model.clj
```diff
@@ -0,0 +1,126 @@
...
+(defn append-count-with-test [model-count test-data n-transitions]
+  (let [test-year (str (dec (Integer/parseInt (:calendar-year model-count))))
+        test-data-for-year  (filter #(= test-year (:calendar-year %)) test-data)
+        send-test-data-for-year  (filter #(not= "NONSEND" (:need-2 %)) test-data-for-year)
+        ground-truth-count (count send-test-data-for-year)]
+    (assoc model-count :ground-truth (str ground-truth-count) :n-transition n-transitions)))
+
+(defn append-state-with-test [model-state test-data n-transitions]
+  (let [test-year (str (dec (Integer/parseInt (:calendar-year model-state))))
+        state-vector (str/split (str/replace (:state model-state) ":" "") #"-")
```
elisehuard: I think you could probably assign it to need and setting and one go using destructing (like in python)
[x,y] (str/split ...)

src/clj/witan/send/validate_model.clj
```diff
@@ -0,0 +1,126 @@
...
+                        (map str))
+        model-state (return-testable-data (str (temp-dir input-path) year "_Output_AY_State.csv") test-years)
+        model-count (return-testable-data (str (temp-dir input-path) year "_Output_Count.csv") test-years)
+        count-results (map #(append-count-with-test % test-data n-transitions) model-count)
+        state-results (map #(append-state-with-test % test-data n-transitions) model-state)]
+    (write-csv (str (temp-dir input-path) "results_" year "_count.csv") count-results)
+    (write-csv (str (temp-dir input-path) "results_" year "_state.csv") state-results)))
+
+(defn validate-fold [input-path year transitions settings]
+  (def temp (temp-dir input-path))
```
acron0: Why is this `def` and not in the `let` below?

mikeyford: good catch, that line shouldn't be there

src/clj/witan/send/validate_model.clj
```diff
@@ -0,0 +1,126 @@
...
+                                          (= academic-year (:academic-year-2 %))) test-data)
+        ground-truth-count (count test-data-for-state)]
+    (assoc model-state :ground-truth (str ground-truth-count) :n-transitions (str n-transitions))))
+
+(defn collate-fold [input-path year n-transitions]
+  (let [test-data (load-csv-as-maps (str (temp-dir input-path) "test_" year ".csv"))
+        test-years (->> (map :calendar-year test-data)
+                        (distinct)
+                        (map #(inc (Integer/parseInt %)))
+                        (into [])
```
acron0: I think this is redundant, at the following `(map str)` will turn it back to lazy seq

src/clj/witan/send/validate_model.clj
```diff
@@ -0,0 +1,126 @@
...
+(defn copy-if-exists [source-path destination-path file]
+  (let [source (str source-path file)]
+    (if (.exists (clojure.java.io/as-file source))
+      (io/copy (io/file source) (io/file (str destination-path file))))))
+
+(defn copy-input-files [input-path year]
+  (let [source-path (str "data/" input-path)
+        destination-path (str (temp-dir input-path) year "/")]
+    (.mkdir (java.io.File. destination-path))
+    (doseq [file ["modify-settings.csv" "population.csv" "need-setting-costs.csv" "valid-setting-academic-years.csv"]]
```
acron0: would be nice if these file names were in a namespace def, rather than in the fn

src/clj/witan/send/validate_model.clj
```diff
@@ -0,0 +1,126 @@
...
+        years-to-validate (get-validation-years transitions)]
+    (doseq [year years-to-validate]
+      (validate-fold input-path year transitions settings-map))
+    (->> (map #(load-csv-as-maps (str "data/" input-path "temp/results_" % "_count.csv")) years-to-validate)
+         (flatten)
+         (write-csv (str "validation_result_count_" (str/join (drop-last input-path)) ".csv")))
+    (->> (map #(load-csv-as-maps (str "data/" input-path "temp/results_" % "_state.csv")) years-to-validate)
+         (flatten)
+         (write-csv (str "validation_result_state_" (str/join (drop-last input-path)) ".csv"))))
+  (if (false? keep-temp-files?)
```
acron0: `(not keep-temp-files?)` is more idiomatic, as it covers `false` and `nil`.

src/clj/witan/send/validate_model.clj
mikeyford: changed

src/clj/witan/send/validate_model.clj
mikeyford: yep, changed

src/clj/witan/send/validate_model.clj
mikeyford: changed

src/clj/witan/send/validate_model.clj
mikeyford: done

PR #51, opened 2018-05-24
comparative charts for alternative scenarios

sunnytownsend:
R code for plotting 2 model run results (such as baseline and scenario) together on the same graph to make it easier to compare outcomes.

Plots included here:
- total cost
- total population count


src/R/comparative-charts.R
```diff
@@ -0,0 +1,45 @@
+library(dplyr)
+library(ggplot2)
+
+
+# DATA
+
+## cost
+baseline_cost_data = "data/total_cost/TH-VersionA-baseline-29-03-18-Cost.csv"
```
seb231: Could we make these args to a function which just point at two directories that contain files ending in `Cost.csv` and `Count.csv`?

src/R/send_cost_plot_comparative.R
```diff
@@ -0,0 +1,56 @@
...
+         height=6,
+         units="in")
+  
+  
+  
+  ## ribbon plot - zero-indexed
+  
+  g + scale_y_continuous(limits = c(0, max(cost_data_baseline$q3, na.rm=T)))
+  
+  ggsave("target/zero_indexed/Total_Cost_Comparative_Zeroindexed.pdf",
```
seb231: remove zero_indexed dir

mikeyford: @seb231 I suggested this this morning in standup, so that we don't have too many plots one one folder and it gets confusing

src/R/population_line_plot_comparative.R
```diff
@@ -0,0 +1,80 @@
...
+    ggtitle("SEND Population Projection") +
+    theme(axis.text.x = element_text(size=8)) +
+    geom_vline(xintercept = max(count_data_historic$calendar.year) + 0.5,
+               color = "dodgerblue",
+               linetype = "dashed")  +
+    geom_text(x=max(count_data_historic$calendar.year) + 0.5,
+              y=max(count_data_baseline$q3, na.rm=T),
+              label = "<-- Historical      Projected -->",
+              color = "dodgerblue") +
+    geom_text(x=2027+1,
```
mikeyford: Is it possible to pull the 2027 from the data, rather than hardcode it? We might not always end projections there.

src/R/send_cost_plot_comparative.R
```diff
@@ -0,0 +1,56 @@
...
+    geom_ribbon(aes(ymin=q1, ymax=q3), fill=cols[1], alpha = alpha_ribbon) +
+    geom_line(data=cost_data_scenario, aes(y=mean), colour=cols[2], alpha = alpha_line) +
+    geom_ribbon(data=cost_data_scenario, aes(ymin=q1, ymax=q3), fill=cols[2], alpha = alpha_ribbon) +
+    labs(y = "Total Projected SEND Cost / £ million") +
+    scale_x_continuous(name="Year",
+                       breaks = seq(min(cost_data_baseline$calendar.year), max(cost_data_baseline$calendar.year)),
+                       limits = c(min(cost_data_baseline$calendar.year), max(cost_data_baseline$calendar.year)+1)) +
+    ggtitle("SEND Cost Projection") +
+    theme(axis.text.x = element_text(size=8)) +
+    geom_text(x=2027+0.8,
```
mikeyford: same here, pull max year rather than hardcode

src/R/population_line_plot_comparative.R
```diff
@@ -0,0 +1,80 @@
...
+              y=count_data_baseline$mean[count_data_baseline$calendar.year==2027]+20,
+              label= "Baseline",
+              colour = cols[1]) +
+    geom_text(x=2027+1,
+              y=count_data_scenario$mean[count_data_scenario$calendar.year==2027]-20,
+              label= "Scenario A",
```
mikeyford: Can you infer "Baseline" and "Scenario A" from the file names maybe? It's unlikely to always be those we want to compare.

src/R/send_cost_plot_comparative.R
seb231: I feel like we're just moving the files to more locations, which is just as confusing

sunnytownsend: removed zero_indexed dir

src/R/comparative-charts.R
sunnytownsend: done

src/R/population_line_plot_comparative.R
sunnytownsend: done

src/R/send_cost_plot_comparative.R
sunnytownsend: 👍

src/R/population_line_plot_comparative.R
sunnytownsend: done - had to make some specific requirements of the folder locations and naming, but hopefully makes sense?

src/R/comparative-charts.R
```diff
@@ -4,13 +4,16 @@ library(ggplot2)
...
+## path to the folder containing the 2 data sets you want to compare. 
+### The code expects the folder and the data to be in specific forms and places: 
+### - the data must be placed inside the data/ folder
+### - the data must be csvs
+### - the data must take the same name as the folder (i.e. TH-VersionA-baseline-29-03-18)
+### - plus an additional name-part for the type of data (e.g. -Cost, -Count) followed by the .csv extension
+### The code will extract what type of run the data comes from (e.g baseline, Scenario 1) the name of the folder, and will use this for
+### labelling the plot
+data1_folder = "data/TH-VersionA-baseline-29-03-18"
+data2_folder = "data/TH-VersionA-scenario1-27-04-18"
```
seb231: So the idea is we just replace the path here?

sunnytownsend: yes (although I will remove `data/` as that is part of the stipulations for use). Is that ok? Any other ideas of doing this better are v welcome

src/R/population_line_plot_comparative.R
sunnytownsend: @mikeyford  at the moment this makes the label Scenario 1 rather than Scenario A - is that correct/ok?

src/R/comparative-charts.R
seb231: I am (honestly) not sure if it makes more sense to "wrap" this so the two folders/sets of data are just args to a script we run from the terminal, or whether it's just easier to open RStudio and change those paths in the code... At the very least I think they should say `path` or something and not refer to the TH data please :)

src/R/population_line_plot_comparative.R
seb231: hmm should be Scenario 1 at least for what we've been testing

seb231: but would be good if we could do as @mikeyford says?

PR #61, opened 2018-07-16
historic R charts use SEND only

mikeyford:


mikeyford:
I did have a look. I think R also filters for SEND only, so it's not made a difference. Probably still worth changing though incase that file is used in a different workflow.

seb231:
that's great thanks 👍

PR #63, opened 2018-07-18
Fix/move sankeys to r

seb231:


src/R/send-charts.R
```diff
@@ -2,7 +2,7 @@ install_missing_packages <- function(package_name) {
   if(!require(package_name, character.only = TRUE)) {
     install.packages(package_name, repos='http://cran.us.r-project.org') } }
 
-packages_to_check = c("dplyr", "ggplot2", "reshape2", "stringr", "devtools", "svglite")
+packages_to_check = c("dplyr", "ggplot2", "reshape2", "stringr", "devtools", "svglite", "ggforce")
```
mikeyford: can we remove the dep on "devtools" (and also remove the commented out code on line 10)?

PR #64, opened 2018-07-19
Fix/move ribbon plots to r

seb231:


src/R/send-charts.R
```diff
@@ -362,6 +362,33 @@ for (s2 in settings) {
...
+ribbon <- function(data, title) {
+  ggplot(data, aes(year, upper0)) +
+    geom_ribbon(aes(ymax = upper0, ymin = lower0, fill = "#1b9e77"), alpha = 0.2) +
+    geom_ribbon(aes(ymax = upper1, ymin = lower1, fill = "#d95f02"), alpha = 0.2) +
+    geom_ribbon(aes(ymax = upper2, ymin = lower2, fill = "#7570b3"), alpha = 0.2) +
+    geom_ribbon(aes(ymax = upper3, ymin = lower3, fill = "#e7298a"), alpha = 0.2) +
+    ggtitle(paste(title, "probability by academic year")) +
+    xlab("NCY") +
+    ylab("95% probability interval") +
+    scale_fill_manual(name = "Years", labels = c("2013", "2014", "2015", "2016"), values = c("#1b9e77", "#d95f02", "#7570b3", "#e7298a")) +
```
mikeyford: I think we probably need to infer this automatically since e.g. the Camden data only has three years of transitions.

mikeyford:
Hey, the *_probability charts don't seem to work for Camden data. Only the joiner_probability is produced but it's blank. I think maybe the number of `geom_ribobn`, colours etc. assume 4 years of transitions and need to be changed aswell maybe?

seb231:
thanks @mikeyford will look into this now

src/clj/witan/send/charts.clj
```diff
@@ -1,159 +0,0 @@
-(ns witan.send.charts
```
mikeyford: oh man...feels good to get rid of this!

mikeyford:
Looks good and tests good locally

PR #67, opened 2018-07-27
Added output file comparision check via md5s, removed bad test

mattford63:
**Added output file comparision check via md5s, removed bad test**


mikeyford:
I did notice that before circle CI did 🥇

PR #68, opened 2018-07-30
main takes config.edn not dir

mikeyford:


mikeyford:
Tests are failing but I'm pretty sure it's a CircleCI github integration issue. Lein test is dieing, not failing. Will try again tomorrow before merging

PR #71, opened 2018-08-06
**Add validation to config file**

mattford63:
Also

- make main only deal with config parsing and selecting what to run
- make validate use the top-level config parsing (removes duplicate code)

### DIFF
--- a/data/demo/config.edn
+++ b/data/demo/config.edn
@@ -17,4 +17,8 @@
  :output-parameters {:run-report-header true
                      :run-reports true
                      :run-charts false
-                     :output-dir "results"}}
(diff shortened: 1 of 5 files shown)


data/demo/config.edn
```diff
@@ -17,4 +17,8 @@
  :output-parameters {:run-report-header true
                      :run-reports true
                      :run-charts false
-                     :output-dir "results"}}
+                     :output-dir "results"}
+
+ :validation-parameters {:keep-temp-files? false
+                         :run-as-default true}
```
seb231: I'm surprised this doesn't break the tests or take a really long time to pass?

PR #74, opened 2018-08-07
Initial reorg

mattford63:
- the names chosen: input, run, and output are not final.

Output feels to do quite a bit at the moment.

There are no tests for run functions.

The swarm will likely further change this.


src/clj/witan/send/send.clj
```diff
@@ -1,659 +1,14 @@
...
-          (with-open [in (io/input-stream (io/resource "send-charts.R"))]
-            (io/copy in (io/file "/tmp/send-charts.R")))
-          (sh/sh "Rscript" "--vanilla" "/tmp/send-charts.R" dir)
-          (run! #(io/delete-file (str dir "/" %) :quiet)
-                ["historic-data.csv" "valid-settings.csv" "joiner-rates.csv"
-                 "leaver-rates.csv" "mover-rates.csv"]))))
-    (report/write-send-report (str dir "/SEND_Log.md"))))
+   (-> (i/build-input-datasets (:project-dir config) (:file-inputs config) (:schema-inputs config))
+       (i/prepare-send-inputs (:transition-parameters config))
+       (r/run-send-model (:run-parameters config)))))
```
seb231: Should this fn just be in main or something, just to dump send.clj? (I understand it may actually be here to avoid circular deps...?)

PR #77, opened 2018-08-09
Instead of copying probability distributions don't provide one

mattford63:
**Instead of copying probability distributions don't provide one.**
This is likely not a full fix.  But does a pointer to the issue that
needs solving.  We can't blindly copy distributions when we have no
data at least for joiners.

### DIFF
--- a/src/clj/witan/send/params.clj
+++ b/src/clj/witan/send/params.clj
@@ -173,7 +173,7 @@
 (defn continue-for-latter-ays [params academic-years]
   (reduce (fn [coll ay]
             (if-let [v (or (get coll ay)
-                           nil)]
+                           nil)] ;; Replace with average of adjacent AY params
               (assoc coll ay v)
               coll))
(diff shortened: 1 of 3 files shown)


seb231:
This is probably an issue that Henry raised if we look carefully through his "future developement" work, but was then lost in the push for alternative scenarios

seb231:
Need to instead of applying no joiner distribution or the distribution from the previous year, need to apply a low-level distribution as we do with unseen transitions

seb231:
I had to fix the checksum tests, as the results are now slightly different with the implemented non-filling of gaps.

The intention is to return to this work in the future, and add a better way of gap filling, but not filling is sufficient for the moment, and more realistic. Gaps are now reported also, we are aware of them.

mikeyford:
Might be worth re-running the model validation with TH data

seb231:
good call

PR #78, opened 2018-08-15
Tidy run fn

mikeyford:


data/demo/config.edn
```diff
@@ -20,5 +20,5 @@
                      :output-dir "results"}
 
  :validation-parameters {:keep-temp-files? false
-                         :run-as-default true}
+                         :run-as-default false}
```
mikeyford: It was annoying me that every time I ran `-main` when working on the model, the validation was running as well.

src/clj/witan/send/model/run.clj
```diff
@@ -53,16 +46,12 @@
...
   Calls 'unsafe' equivalent once we've removed non-send and children outside
   valid academic year range."
-  [[model transitions :as model-state] [[year state] population :as cohort]
-   {:keys [joiner-beta-params joiner-state-alphas
-           leaver-beta-params
-           mover-beta-params mover-state-alphas
-           valid-year-settings] :as params}
-   calendar-year]
+  [[model transitions :as model-state]
+   [[year state] population :as cohort]
```
mikeyford: will fix the alignment here

PR #87, opened 2018-09-20
Feature/allow settings exclusion in charts

seb231:


fhr:
I've reviewed the R and happy it does what's intended (note that "Elsewhere" now doesn't get explicitly excluded unless passed to the args)
No comment on the Clojure

PR #94, opened 2018-10-19
Removes redundant code

otfrom:
A fair bit of redundant code has grown in this repo. This attempts to remove most to all of it.

There is one place where some code is moved to a test namespace where it is used.


project.clj
```diff
@@ -31,7 +31,8 @@
                  [org.clojure/math.combinatorics "0.1.4"]]
   :plugins [[lein-gorilla "0.4.0"]
             [lein-cljsbuild "1.1.4"]
-            [cider/cider-nrepl "0.18.0-SNAPSHOT"]]
+            [cider/cider-nrepl "0.18.0-SNAPSHOT"]
+            [lein-cloverage "1.0.13"]]
```
seb231: Do we need to push a version cloverage in it? Surely it's just for testing coverage locally and we won't need it (for the most part after this round?)

seb231: Do we need to push a version cloverage in it? Surely it's just for testing coverage locally and we won't need it (for the most part after this round?)

seb231: Do we need to push a version cloverage in it? Surely it's just for testing coverage locally and we won't need it (for the most part after this round?)

seb231: Do we need to push a version cloverage in it? Surely it's just for testing coverage locally and we won't need it (for the most part after this round?)

PR #100, opened 2018-10-30
Tidied testing statement.

otfrom:


test/witan/send/acceptance/validate_model_test.clj
```diff
@@ -1,9 +1,10 @@
 (ns witan.send.acceptance.validate-model-test
-  (:require [clojure.test :refer :all]
-            [clojure.java.io :as io]
+  (:require [clojure.java.io :as io]
             [clojure.string :refer [join]]
+            [clojure.test :refer [deftest is testing]]
```
seb231: What does this mean?

otfrom: It just explicitly refers things in rather than a refer all. I'm deciding if I prefer it or not. It does have the advantage of being specific and keeping the joker linter quiet. :smile_cat:

seb231: I mean `[deftest is testing]`? What does "is testing" mean?

otfrom: `deftest`, `is`, and `testing` are the 3 macros that are being referred to, so there is no `is testing` as such.

seb231:
Ah! Thank you!

PR #104, opened 2018-11-09
Split csv loading from prep for model run

otfrom:


src/clj/witan/send/model/prepare.clj
```diff
@@ -0,0 +1,217 @@
...
+            [medley.core :as medley]
+            [witan.send.params :as p]
+            [witan.send.check-inputs :refer [run-input-checks]]
+            [clojure.set :refer [rename-keys]]
+            [clojure.walk :refer [postwalk]]
+            [witan.send.report :as report]
+            [schema.core :as s]
+            [witan.send.schemas :as sc]))
+
+(defn int-ceil [n]
```
seb231: Still not getting rid of this or building into where it is used?

otfrom: just this mechanical bit as a first pass. We an kill it off later when we look at it. (iterateiterateiterate)

PR #107, opened 2018-11-13
Doc string for weighted joiner beta params

mattford63:


src/clj/witan/send/params.clj
```diff
@@ -196,7 +198,7 @@
     (continue-for-latter-ays params academic-years)))
 
 (defn beta-params-leavers [valid-states transitions]
-  #_(weighted-beta-params valid-states transitions joiner? leaver?)
+  ;;https://docs.google.com/document/d/1fmo3ecHdRNnEECLCTdqie5QJ-COvbIBBBghCSAXlHN0/edit(weighted-beta-params valid-states transitions joiner? leaver?)
```
otfrom: this doesn't belong here.

PR #108, opened 2018-11-15
Fix/better config parameter naming

seb231:


data/demo/config.edn
```diff
@@ -1,18 +1,16 @@
-{:file-inputs {:settings-to-change "data/modify-settings1.csv"
-               :transition-matrix "data/transitions.csv"
+{:file-inputs {:transitions "data/transitions.csv"
                :population "data/population.csv"
-               :setting-cost "data/need-setting-costs.csv"
-               :valid-setting-academic-years "data/valid-setting-academic-years.csv"}
+               :costs "data/need-setting-costs.csv"
+               :valid-states "data/valid-setting-academic-years.csv"}
```
mikeyford: worth maybe rename the demo data too?

seb231: ok

PR #114, opened 2018-11-21
seperate alpha params into own function, replace dup check with sets,…

mikeyford:
… don't include origin setting in prior calc from begining

### DIFF
--- a/src/clj/witan/send/params.clj
+++ b/src/clj/witan/send/params.clj
@@ -1,15 +1,7 @@
 (ns witan.send.params
   (:require [witan.send.constants :as c]
             [witan.send.maths :as m]
-            [witan.send.states :as s]
-            [witan.send.utils :as u]))
-
-(defn- find-duplicates
-  "Finds values that are duplicated in the seq"
-  [seq]
-  (for [[id freq] (frequencies seq)
-        :when (> freq 1)]
-    id))
+            [witan.send.states :as s]))
 
 (defn joiner?
   [[ay state-1 state-2]]
@@ -187,6 +179,33 @@
     (continue-for-latter-ays params academic-years)))
 
 
+(defn mover-alpha-priors
+  "Calculate uniform prior across all valid transitions from state"
+  [need-setting valid-transitions valid-settings observations-per-ay ay]
+  (let [[need setting] (s/split-need-setting need-setting)
+        allowed-settings (clojure.set/intersection (set (get valid-settings [ay need]))
+                                                   (set (get valid-transitions setting))
+                              ;(remove #{setting})
+                          )
+        prior (->> (zipmap allowed-settings (repeat (/ 1.0 (count allowed-settings))))
+                   (merge-with + (get observations-per-ay ay))
+                   (#(dissoc % setting)))
+        total (->> (vals prior) (apply +))
+        _ (println "new total" total)]
+    (reduce (fn [coll [setting v]]
+              (assoc coll (s/join-need-setting need setting) (double (/ v total))))
+            {}
+            prior)
+    #_(zipmap (map #(s/join-need-setting need %) (keys prior)) (vals prior))))
+
+
+
+(defn- find-duplicates
+  "Finds values that are duplicated in the seq"
+  [seq]
+  (for [[id freq] (frequencies seq)
+        :when (> freq 1)]
+    id))
 
 (defn alpha-params-movers
   "calculates the rate of transitions to a new state at academic year X for state Y"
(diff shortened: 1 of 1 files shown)


src/clj/witan/send/params.clj
```diff
@@ -187,6 +180,18 @@
     (continue-for-latter-ays params academic-years)))
 
 
+(defn mover-alpha-priors
+  "Calculate uniform prior across all valid transitions from state"
```
mattford63: it's not uniform in general

mattford63: it's not from state but rather using observed settings per ay

src/clj/witan/send/params.clj
```diff
@@ -187,6 +180,18 @@
...
+  [need-setting valid-transitions valid-settings observations-per-ay ay]
+  (let [[need setting] (s/split-need-setting need-setting)
+        allowed-settings (->> (intersection (set (get valid-settings [ay need]))
+                                            (set (get valid-transitions setting)))
+                              (remove #{setting}))
+        prior (->> (zipmap allowed-settings (repeat (/ 1.0 (count allowed-settings))))
+                   (merge-with + (get observations-per-ay ay)))]
+    (zipmap (map #(s/join-need-setting need %) (keys prior)) (vals prior))))
```
mattford63: This looks to be missing the normalisation ie. the the division through of the counts and priors by the total.

src/clj/witan/send/params.clj
```diff
@@ -187,6 +180,18 @@
     (continue-for-latter-ays params academic-years)))
 
 
+(defn mover-alpha-priors
+  "Calculate uniform prior across all valid transitions from state"
+  [need-setting valid-transitions valid-settings observations-per-ay ay]
+  (let [[need setting] (s/split-need-setting need-setting)
+        allowed-settings (->> (intersection (set (get valid-settings [ay need]))
```
mattford63: much cleaner

src/clj/witan/send/params.clj
mattford63: I think this would be a good place to write some tests around the original function and prove out our refactor doesn't change the result.  Tests around the whole function to start with and more specific tests around the prior calculation.

mattford63: As to whether it should be normalised is another question entirely....but we have been to date.

mattford63: My guess is this is also why the validation broke.

src/clj/witan/send/params.clj
```diff
@@ -2,14 +2,7 @@
...
             [witan.send.states :as s]
-            [witan.send.utils :as u]))
-
-(defn- find-duplicates
-  "Finds values that are duplicated in the seq"
-  [seq]
-  (for [[id freq] (frequencies seq)
-        :when (> freq 1)]
-    id))
+            [clojure.set :refer [intersection]]))
```
otfrom: I prefer to avoid refer (except in test namespaces)

src/clj/witan/send/params.clj
otfrom: :+1:

src/clj/witan/send/params.clj
```diff
@@ -205,7 +210,7 @@
                                                 (s/join-need-setting need-2 setting-2)] m/some+ 1))
                              {} mover-transitions)
 
-        observations-per-ay (reduce (fn [coll {:keys [academic-year-1 need-1 setting-1 need-2 setting-2]}]
+        observations-per-ay (reduce (fn [coll {:keys [academic-year-1 _ _ _ setting-2]}]
```
otfrom: only pull out the keys you need. You don't need the `_ _ _`

src/clj/witan/send/params.clj
```diff
@@ -216,24 +221,9 @@
...
-                    prior (->> (zipmap settings (repeat (/ 1.0 (count settings))))
-                               (merge-with + ay-obs))
-                    prior (dissoc prior setting)
-                    total (->> (vals prior) (apply +))
-                    prior (reduce (fn [coll [setting v]]
-                                    (assoc coll (s/join-need-setting need setting) (double (/ v total))))
-                                  {}
-                                  prior)]
-                (assoc coll [ay state] (merge-with + prior obs))))
+    (reduce (fn [coll [ay need-setting]]
```
otfrom: I wonder about moving this fn and its helpers to a separate namespace. wdyt?

src/clj/witan/send/params.clj
```diff
@@ -187,14 +179,41 @@
...
 
+(defn- mover-alpha-prior
+  "Calculate prior by adding uniform distribution across allowed settings to observations and normalising."
+  [need-setting valid-transitions valid-settings observations-per-ay ay]
+  (let [[need setting] (s/split-need-setting need-setting)
+        allowed-settings (clojure.set/intersection (set (get valid-settings [ay need]))
+                                                   (set (get valid-transitions setting)))
+        prior (->> (zipmap allowed-settings (repeat (/ 1.0 (count allowed-settings))))
+                   (merge-with + (get observations-per-ay ay))
+                   (#(dissoc % setting)))
```
otfrom: this can be rewritten as:

```clojure
  (-> allowed-settings
      (zipmap (repeat (/ 1.0 (count allowed-settings))))
      ((partial merge-with + (get observations-per-ay ay)))
      (dissoc setting))
```

merge-with is the only thing that annoyingly doesn't play nicely with the `->`

PR #115, opened 2018-11-22
Remove try except

mattford63:
**Remove try except**

**Rename probs**

**start to rename distributions appropriately**

**rename symbols in sample-send-transitions**

**Keyword arguments for run functions**

**state renaming , pop func renaming**

### DIFF
--- a/src/clj/witan/send/model/run.clj
+++ b/src/clj/witan/send/model/run.clj
@@ -9,21 +9,23 @@
             [witan.send.step :as step]
             [witan.send.utils :as u]))
 
-(defn incorporate-new-states-for-academic-year-state
+(defn incorporate-new-ay-need-setting-populations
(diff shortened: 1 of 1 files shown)


src/clj/witan/send/model/run.clj
```diff
@@ -9,54 +9,82 @@
...
-        (-> (d/sample-dirichlet-multinomial movers probs)
-            (assoc state non-movers)))
-      {})
-    (catch Exception e
-      (do (println state n probs mover-beta)
-          nil))))
+               (update [calendar-year academic-year need-setting next-need-setting] m/some+ n)))
+           transitions predicted-populations)))
+
+(defn predict-movers
```
mikeyford: this is a much better name and useful docstring

PR #119, opened 2018-11-28
Fixed case for needs plots dependent on input

seb231:


seb231:
Agreed, pretty sure there's a card in their somewhere for that

PR #127, opened 2018-12-19
1/2 joiners for ay>12

mattford63:


seb231:
@mattford63 I tried following the chat around this PR in Slack, but I'm still not really sure what's going on. Would you be able to provide a description for this pls?

PR #130, opened 2018-12-30
Make chart generation OS independent

mattford63:
- discover os tmp location
- add Win 10 Rscript location


mikeyford:
works fine locally for me still on on me apple$$

PR #131, opened 2019-01-03
TBD: No copy aging

mattford63:
Hide the white-space changes before review (they are slight but annoying)

Instead of copying the population state for each entity (need, setting, ay) to (need,setting,ay +1) in the model and then subsequently modifying it once we work out what the new populations should be, we simply work out the new population and place this result in (need,setting, ay +1).

This removes the possibility that if any entity for whatever reason is not modified by the new population calculation having its population propagated (i.e prevents it acting like a remainer/stayer).

If the two methods were equivalent then the impact would be zero and the validation would pass.  However validation does not pass.  Inspection of the visualisation over the demo data and Tower Hamlets data shows that the results are altered only slightly.  This suggests that there are at least some cases of unintended remainers.


src/clj/witan/send/model/run.clj
```diff
@@ -161,24 +164,24 @@
...
-                                                                                                       u/model-population-by-need-setting))
+                  (r/fuse {:by-state                  (u/model-states-rf validate-valid-states (u/histogram-rf number-of-significant-digits))
+                           :total-in-send-by-ay       (r/pre-step (u/with-keys-rf (u/histogram-rf number-of-significant-digits) sc/academic-years) u/model-population-by-ay)
+                           :total-in-send             (r/pre-step (u/histogram-rf number-of-significant-digits) u/model-send-population)
+                           :total-in-send-by-need     (r/pre-step (u/merge-with-rf (u/histogram-rf number-of-significant-digits)) u/model-population-by-need)
+                           :total-in-send-by-setting  (r/pre-step (u/merge-with-rf (u/histogram-rf number-of-significant-digits)) u/model-population-by-setting)
+                           :total-cost                (r/pre-step (u/histogram-rf number-of-significant-digits) (comp (partial u/total-need-setting-cost cost-lookup)
```
seb231: \w :face_vomiting:

mattford63: I put this back to the original spacing

PR #137, opened 2019-01-10
state-pop-output-only + rate fix

mattford63:


src/clj/witan/send/maths.clj
```diff
@@ -8,3 +8,8 @@
   "This rounds numbers to 2 decimal places."
   [x]
   (Double/parseDouble (format "%.02f" (double x))))
+
+(defn round0
+  "This rounds numbers to 2 decimal places."
```
mikeyford: this should be 0 right?

PR #139, opened 2019-01-11
Feature/throw exception if input problem

seb231:


src/clj/witan/send/send.clj
```diff
@@ -10,9 +10,10 @@
   generate it)"
   ([config]
    (reset-send-report)
-   (-> (i/build-input-datasets (:project-dir config) (:file-inputs config) (:schema-inputs config))
-       (p/prepare-send-inputs (:transition-parameters config))
-       (r/run-send-model (:projection-parameters config)))))
+   (let [input (i/build-input-datasets (:project-dir config) (:file-inputs config) (:schema-inputs config))]
+     (-> (try (p/prepare-send-inputs input (:transition-parameters config))
+              (catch Exception e (println (clojure.string/join "\n" (i/check-if-dataset-is-valid input)))))
```
seb231: @otfrom the exception isn't being caught here (this is the problem I had yesterday evening). `prepare-send-inputs` is failing at the first fn it trys to call (`row-maps`) and throwing a fairly meaningless exception, which is why I'm not capturing the message, but I instead want it to do `check-if-dataset-is-valid` in that instance, and then give me back a message with the problem

seb231:
I found a third time of error, which was much harder to differeniate between whether the input data was missing or incorrect so needed to simplify the checks. I think this is fine, as ultimately what you need to do is 1. Check the config file is correct and/or then 2. Check what the data in the offending files looks like

src/clj/witan/send/model/input.clj
```diff
@@ -47,6 +47,24 @@
       (apply-schema-coercion schema)
       (as-> {:keys [column-names columns]} (ds/dataset column-names columns))))
 
+(defn is-dataset-erroneous? [dataset]
+  (->> dataset second :column-names first ((complement instance?) clojure.lang.Keyword)))
+
+(defn check-dataset [pred s datasets]
+  (when (some true? (map pred datasets))
```
mattford63: Do you need this given you filter on the predicate anyway?

seb231: Yep you're right, not needed now it's a single pred

PR #147, opened 2019-01-24
chart-tidy-up

mattford63:
You may need to run twice to install the optparse library.

### DIFF
--- a/src/clj/witan/send/model/output.clj
+++ b/src/clj/witan/send/model/output.clj
@@ -16,7 +16,7 @@
 
 (defn confidence-bounds
   [results calendar-year]
-  (let [academic-years (keys results)]
+  (if-let [academic-years (keys results)]
     (for [academic-year (sort academic-years)]
       (let [alpha (get-in results [academic-year calendar-year :alpha] 0)
             beta (get-in results [academic-year calendar-year :beta])]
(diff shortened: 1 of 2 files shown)


src/clj/witan/send/model/output.clj
```diff
@@ -16,7 +16,7 @@
 
 (defn confidence-bounds
   [results calendar-year]
-  (let [academic-years (keys results)]
+  (if-let [academic-years (keys results)]
```
otfrom: should this be a `when-let`?

src/clj/witan/send/model/output.clj
```diff
@@ -48,15 +48,16 @@
...
-  (let [n-years (count years)
-        filter-data (map #(pull-year data %) (range n-years))
-        ay (first (first filter-data))
-        uppers (create-CI-map "upper" filter-data 1 n-years)
-        lowers (create-CI-map "lower" filter-data 2 n-years)
-        non-zero-data (filter valid-years-vector? (first data))
-        min-year (first (first non-zero-data))
-        max-year (first (last non-zero-data))]
-    (merge lowers uppers {:year ay})))
+  (if (not (nil? (first data)))
```
otfrom: I think this could be `(when (first data)` unless there is a second leg I don't see. You can also nil pun directly.

mattford63:
> Any ideas why optparse does't install and run on a single run through?

It doesn't seem to wait for the install (or dynamically load it into the existing console).  It seems a problem for all these libraries.  Not a fix to worry about I think.

seb231:
Rebase please :)

PR #151, opened 2019-01-28
add count of missing costs in transitions to report

mikeyford:
This extends the missing cost input check to count the number of times each missing costs occurs in the transitions input. This is useful to feedback to clients where they should focus on if they need to fix costs, and also as a warning to us if a common entity is missing a cost (and will be used as zero)

### DIFF
--- a/src/clj/witan/send/check_inputs.clj
+++ b/src/clj/witan/send/check_inputs.clj
@@ -67,17 +67,25 @@
        (remove (fn [t] (= (+ 1 (:academic-year-1 t)) (:academic-year-2 t))))
        (map #(str "Academic years 1 and 2 are not incremental for " %))))
 
-(defn cost-for-need-setting? [state costs]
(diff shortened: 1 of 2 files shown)


src/clj/witan/send/check_inputs.clj
```diff
@@ -69,15 +69,22 @@
 
 (defn cost-for-need-setting? [state costs]
   "Takes map containing state and returns true for match in costs"
-  (some? (some true? (map #(and (= (:need state) (:need %))
-                                (= (:setting state) (:setting %))) costs))))
+  (some? (some true? (map #(and (= (first state) (:need %))
```
mattford63: Why've you changed the explicit lookup - this is less clear to me.

mattford63: Also perhaps change `state` to `need-setting`?

mattford63: cool

PR #152, opened 2019-02-01
Output the initial population starting state for the simulation

mattford63:


seb231:
What's the purpose of this?

PR #155, opened 2019-02-18
Conditionally output scenarios

mattford63:


seb231:
@mattford63 sorry I know this has already been merged, but could you add details on how to use this feature? [Here](https://github.com/MastodonC/witan.send/blob/master/doc/config.md) there is a doc with description of all key parameters available in a config file, would you be able to add to is please?

seb231:
@mattford63 sorry I know this has already been merged, but could you add details on how to use this feature? [Here](https://github.com/MastodonC/witan.send/blob/master/doc/config.md) there is a doc with description of all key parameters available in a config file, would you be able to add to is please?

PR #158, opened 2019-02-21
Fix transitions and ci

mikeyford:
The new data was created from the old by 

- recoding existing settings and needs to new alphabet characters (e.g. settings now take values 'A' to 'S', and needs 'T' to 'Y'. 
- sampling the 95% rows randomly with replacement and shuffling the order.
- and adjusting 20% of the row's AYs down a year, and 20% up a year.


data/demo/results_for_checking/SEND_Log.md
```diff
@@ -0,0 +1,19 @@
+__There are no leavers from AY 20__
```
otfrom: Are you happy with these warnings?

data/demo/results_for_checking-spliced/SEND_Log.md
```diff
@@ -0,0 +1,19 @@
+__There are no leavers from AY 20__
```
otfrom: Are you happy with these warnings?

mikeyford: Yeah, no issues around these. (The previous data had the same ones plus a lot more: https://github.com/MastodonC/witan.send/blob/master/data/demo/results/SEND_Log.md)

data/demo/results_for_checking/SEND_Log.md
mikeyford: (as below)

seb231:
"    * and adjusting 20% of the row's AYs down a year, and 20% up a year."
Did you do this for each calendar year? Does this mean that the populations are now unequal?

data/demo/data/costs.csv
```diff
@@ -1,109 +1,89 @@
...
-UKN,MSSOB,0
-UKN,MSSR,0
-UKN,MU,0
-UKN,MUOB,0
-UKN,NMSS,0
-UKN,NMSSR,0
-UKN,OOE,0
\ No newline at end of file
+need,setting,cost
+U,G,1000.0
```
seb231: Does it make sense to at least introduce some variation in cost?

data/demo/data/modify-settings1.csv
```diff
@@ -1,3 +0,0 @@
-setting-1,setting-2
```
seb231: if you remove this file and modify-settings2.csv then you will need to update where [this doc](https://github.com/MastodonC/witan.send/blob/master/doc/scenarios.md) refers to

data/demo/data/valid-states.csv
```diff
@@ -1,21 +1,20 @@
...
-MSSOP,Special,-2,15,"CI,CL,OTH,SEMH,SP,UKN",MSSOP
-MSSR,Special,5,15,"CI,CL,OTH,SEMH,SP,UKN","EO,EYE,FEC,IMS,ISS,ISSR,MAP,MMS,MMSIB,MMSOB,MSS,MSSIB,MSSOB,MSSR,MU,MUOB,NMSS,NMSSR,OOE"
-MU,Other,-2,14,"CI,CL,OTH,SEMH,SP,UKN","EO,EYE,FEC,IMS,ISS,ISSR,MAP,MMS,MMSIB,MMSOB,MSS,MSSIB,MSSOB,MSSR,MU,MUOB,NMSS,NMSSR,OOE"
-MUOB,Other,-2,14,"CI,CL,OTH,SEMH,SP,UKN","EO,EYE,FEC,IMS,ISS,ISSR,MAP,MMS,MMSIB,MMSOB,MSS,MSSIB,MSSOB,MSSR,MU,MUOB,NMSS,NMSSR,OOE"
-NMSS,Special,-2,15,"CI,CL,OTH,SEMH,SP,UKN","EO,EYE,FEC,IMS,ISS,ISSR,MAP,MMS,MMSIB,MMSOB,MSS,MSSIB,MSSOB,MSSR,MU,MUOB,NMSS,NMSSR,OOE"
-NMSSR,Special,-2,15,"CI,CL,OTH,SEMH,SP,UKN","EO,EYE,FEC,IMS,ISS,ISSR,MAP,MMS,MMSIB,MMSOB,MSS,MSSIB,MSSOB,MSSR,MU,MUOB,NMSS,NMSSR,OOE"
-OOE,Other,-4,21,"CI,CL,OTH,SEMH,SP,UKN","EO,EYE,FEC,IMS,ISS,ISSR,MAP,MMS,MMSIB,MMSOB,MSS,MSSIB,MSSOB,MSSR,MU,MUOB,NMSS,NMSSR,OOE"
\ No newline at end of file
+setting,setting-group,min-academic-year,max-academic-year,needs,setting->setting
+A,group1,-3,15,"T,U,V,W,X,Y","A,B,C,D,E,F,G,H,I,J,K,L,N,O,P,Q,R,S"
```
seb231: This looks like all needs and settings transitions are possible, but where did your AYs come from? Does it match the "historic" data?

data/demo/data/costs.csv
mikeyford: would be fairly easy to adjust a few at random...what are you thinking would be the advantage with it though?

data/demo/data/modify-settings1.csv
mikeyford: ok, I'll add these files in again with new setting IDs.

data/demo/data/valid-states.csv
mikeyford: this is just the mins and maxes from each setting in the new data

data/demo/data/modify-settings1.csv
mikeyford: I'm going to add these with a seperate PR @seb231

PR #161, opened 2019-02-25
Scenario Example: double_joiner_a

mattford63:
Example of doubling a joiner population by 2

- this is hardcoded to run on https://github.com/MastodonC/witan.send-runs/tree/master/scenario/data

- it achieves the same as https://github.com/MastodonC/witan.send-runs/blob/master/scenario/config-joiners_a_2.edn

- it is just to show where hooks into the code may go.

- it highlights that there is two transition formats in use by the code base, this code would be simpler if this was refactored slighlty.

### DIFF
--- a/src/clj/witan/send/params.clj
+++ b/src/clj/witan/send/params.clj
@@ -43,9 +43,22 @@
               coll))
           params (sort academic-years)))
 
+(defn modify-transitions-double-joiner-a
+  ; It's unfortunate that the transition format differs between the distribution calculation functions calls.
+  ; Some use raw transitions data, some use transition data having had transition-map called over it.  This means we
+  ; need two versions of this functionality.
+  "Double transitions for setting-A"
+  [transitions]
+  (flatten (for [t transitions]
+             (if (= (:setting-2 t) :A)
+               (repeat 2 t)
+               t))))
+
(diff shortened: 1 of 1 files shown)


src/clj/witan/send/params.clj
```diff
@@ -43,9 +43,22 @@
...
+  (flatten (for [t transitions]
+             (if (= (:setting-2 t) :A)
+               (repeat 2 t)
+               t))))
+
 (defn calculate-joiners-per-calendar-year
   [transitions]
   (->> (filter transitions-joiner? transitions)
+       ; this modification could be any suitable function
+       (modify-transitions-double-joiner-a)
```
seb231: Surely this just increases the number of joiners per calendar year. If we want to keep the same amount of students joining SEND each year, but make it "more likely" that they will join a specific setting, we just have to use `modify-transitions-double-joiner-a-transition-map` where the alpha is affected

mattford63: Yep, but this example was to show how to mimic the existing scenario functionality.  And also highlight how we can, if we choose, alter things in a decoupled way i.e only use `modify-transitions-double-joiner-a-transition-map`.

seb231: ah great! 👍

PR #165, opened 2019-03-01
Output transitions branch with white space changes nuked

mattford63:
Here's a summary of the changes I've made needed to get a very small transitions data set producing sensible results. The impact on client data runs seems small.

I'd like to go over these changes and get agreement that the results are correct and then establish unit tests for the various distro calculations and acceptance tests over some small data sets.

* adds output transition data as a CSV
* changes transition data to use the ay "from" in the transition output (for consistency with input transitions)
* remainers that would have aged to invalid group and then be removed via the cleanup are now explicitly prevented and added to the transitions file as leavers
* fixed a joiner beta calculation bug in which no observed joiners also means we also don't count the number of non-joiners for a calendar year
* added unobserved priors for mover beta distribution (rather than fall back to kixi.stats default)
* added unobserved priors for leaver beta distribution (rather than fall back to kixi.stats default)
* added needs to the mover dirichlet prior calculation
* fixed mover distribution calculations to allow the possibility of transitioning between need within the same setting.

### DIFF
--- a/src/clj/witan/send/distributions.clj
+++ b/src/clj/witan/send/distributions.clj
@@ -17,6 +17,7 @@
 
 (defn sample-beta-binomial
   [n params]
-  (if (pos? n)
+  (if (and (pos? n)
+           (not= params {:alpha 0 :beta 0}))
     (d/draw (d/beta-binomial n params) {:seed (get-seed!)})
     0))
--- a/src/clj/witan/send/params.clj
+++ b/src/clj/witan/send/params.clj
@@ -35,20 +35,17 @@
   [[ay _ _]]
   ay)
 
-(defn continue-for-latter-ays [params academic-years]
-  (reduce (fn [coll ay]
-            (if-let [v (or (get coll ay)
-                           nil)] ;; Replace with average of adjacent AY params
-              (assoc coll ay v)
-              coll))
-          params (sort academic-years)))
-
 (defn calculate-joiners-per-calendar-year
+  "The result maps keys are used to lookup up external population so need to be added
+   even though there is no observed count"
   [transitions]
+  (let [observed-calendar-years (into #{} (map :calendar-year transitions))
+        joiners-per-calendar-year (into {} (for [cy observed-calendar-years] [cy {}]))]
   (->> (filter transitions-joiner? transitions)
        (reduce (fn [coll {:keys [calendar-year academic-year-2]}]
                  (update-in coll [calendar-year academic-year-2] m/some+ 1))
-               {})))
+                 {})
+         (merge joiners-per-calendar-year))))
 
 (defn calculate-population-per-calendar-year
   [population]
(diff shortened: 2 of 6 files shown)


src/clj/witan/send/distributions.clj
```diff
@@ -17,6 +17,7 @@
 
 (defn sample-beta-binomial
   [n params]
-  (if (pos? n)
+  (if (and (pos? n)
+           (not= params {:alpha 0 :beta 0}))
```
seb231: this is instead of the kixi.stats `{:alpha 1 :beta 1}` right?

seb231:
What does `with white space changes nuked` mean?

src/clj/witan/send/model/output.clj
```diff
@@ -85,8 +85,24 @@
                                    :beta (- (get-in population [calendar-year academic-year]) j)})))
                     coll years)) {} ages))
 
-(defn output-transitions [file projections]
-  (binding [*print-length* nil] (spit file (pr-str projections))))
+(defn output-transitions
```
seb231: Do we need to do both of these outputs now? Also it seems a bit wrong using the same defn for a function which does two quite different things

seb231: docstring would be cool too

src/clj/witan/send/params.clj
```diff
@@ -108,45 +106,53 @@
...
-                                                 (get-in observations [ay state])
-                                                 (get prior-per-year ay))]
-                  (assoc coll [ay state] beta-params)
+                                                   (get-in observations [ay need-setting])
+                                                   (get prior-per-year ay)
+                                                   (get unobserved-priors ay))]
+                    (assoc coll [ay need-setting] beta-params)
                   coll)
-                coll))
+                  coll)
```
seb231: remove return line

seb231:
@mattford63 not sure if you did a squash but would have been easier to follow if the changes were attached to commits similar to your list above. Just an observation

src/clj/witan/send/params.clj
```diff
@@ -165,33 +171,49 @@
...
+                                ; no observed prior
+                                (assoc coll ay prior-alphas))))
                        {}
-                       academic-years)]
-    (continue-for-latter-ays params academic-years)))
+                          academic-years)
+        dirichlet-parameter-space-size (reduce (fn [c row] (+ c (count (second row)))) 0 dirichlet)]
+    (println (str "\nnumber-of-unique-joiner-transitions: " number-of-unique-joiner-transitions))
+    (println (str "joiner-dirichlet-parameter-space-size:" dirichlet-parameter-space-size))
+    (println (str "joiner-percentage-of-dirichlet-using-only-priors: " (- 100 (double (* 100 (/ number-of-unique-joiner-transitions dirichlet-parameter-space-size))))))
```
seb231: Would it make more sense for these strings to have spaces rather than hyphens?

src/clj/witan/send/params.clj
```diff
@@ -165,33 +171,49 @@
...
+    dirichlet))
 
+(defn cart [colls]
```
seb231: what does this do?

mattford63:
Nuked white space means I just got rid of the whitespace changes so that it
would be easier to split up.

PR #166, opened 2019-03-07
total general population chart for comparison

mikeyford:
Chart for total general population in LA over historic and projected period, requested by Brent.

image

### DIFF
--- a/resources/send-charts.R
+++ b/resources/send-charts.R
@@ -104,7 +104,7 @@ ggplot(df_ay, aes(x=calendar.year, y=value, group=variable)) +
   scale_y_continuous(name='SEND Population') +
   scale_color_manual(name = "NCY",
                      values = c("cyan", "coral", "dodgerblue", "hotpink", "green4"),
-                     labels = c("0 and under", "1 to 6", "7 to 11", "12 to 13", "14 and up")) +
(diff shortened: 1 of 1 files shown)


resources/send-charts.R
```diff
@@ -104,7 +104,7 @@ ggplot(df_ay, aes(x=calendar.year, y=value, group=variable)) +
   scale_y_continuous(name='SEND Population') +
   scale_color_manual(name = "NCY",
                      values = c("cyan", "coral", "dodgerblue", "hotpink", "green4"),
-                     labels = c("0 and under", "1 to 6", "7 to 11", "12 to 13", "14 and up")) +
+                     labels = c("early years", "1 to 6", "7 to 11", "12 to 13", "14 and up")) +
```
seb231: "Early"?

mattford63:
I'd rather this wasn't in the model as it doesn't seem to rely on any model outputs.  But for ease I'm gonna approve :-)

seb231:
> I'd rather this wasn't in the model as it doesn't seem to rely on any model outputs. But for ease I'm gonna approve :-)

Yeah agreed, I think we just need a card in the backlog to move a load of this stuff to the eda

PR #168, opened 2019-03-07
CSV version of Transition Output

mattford63:
https://www.pivotaltracker.com/story/show/164468840

### DIFF
--- a/src/clj/witan/send/model/output.clj
+++ b/src/clj/witan/send/model/output.clj
@@ -85,24 +85,26 @@
                                    :beta (- (get-in population [calendar-year academic-year]) j)})))
                     coll years)) {} ages))
 
-(defn output-transitions
-  ([dir filename projections]
-   (binding [*print-length* nil] (spit (str dir "/" filename ".edn") (pr-str projections))))
-  ([dir filename projections simulations]
-   (with-open [writer (io/writer (io/file (str dir "/" filename ".csv")))]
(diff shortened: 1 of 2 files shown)


src/clj/witan/send/model/output.clj
```diff
@@ -85,8 +85,24 @@
...
+           rows (sort (remove empty?
+                              (for [[[cy ay need-setting-source need-setting-destination] transitions] projections]
+                                (if (> (m/round0 (/ transitions simulations)) 0)
+                                  (into [] (flatten [cy
+                                                     ay
+                                                     (parse-need-setting need-setting-source)
+                                                     (inc ay)
+                                                     (parse-need-setting need-setting-destination)
+                                                     (m/round0 (/ transitions simulations))]))))))]
+       (csv/write-csv writer (into [headers] rows))))))
```
seb231: Do we need the original one of these still? the .edn version?

mattford63: Can't hurt to have it.  Will make a handy comparison if we wanted to test.  Neither are completely accurate at the moment but those fixes are to come.

seb231: Not accurate, in that they don't capture everything?

mattford63: And too much!  And oddly.

mattford63: Should all be in the blogs/other big PR.

seb231: okay. I'm not sure we need both, but it doesn't really hurt. I don't like however that the same fn does two very different things if you use 3 or 4 args. Could you separate them and call them different things please?

PR #170, opened 2019-03-10
Refactor acceptance testing

mattford63:
The goal being to be able to simply replace the _for_checking
directories with files that are approved for validation checking.
i.e, we can simply copy our approved results into these directories
and acceptance testing will then pass.  Very handy for the upcoming
tech debt changes that will alter our results.

- removed all hardcoded md5sums
- removed plot existence checking (not actually used - it was
  previously commented out, and was not a proper test)
- added config files for the _for_checking_ results directory
- abstracted useful acceptance test fns into acceptance.utils ns.


test/witan/send/acceptance/utils.clj
```diff
@@ -0,0 +1,37 @@
...
+            [clojure.java.io :as io]
+            [witan.send.validate-model :as vm]
+            [witan.send.model.input :as i]
+            [clojure.core.matrix.dataset :as ds]
+            digest))
+
+(defn get-md5s [config files]
+  (if-let [output-dir (m/get-output-dir config)]
+    (into {} (for [f files]
+               [f (-> (io/file output-dir f) (digest/md5))]))))
```
otfrom: I'd prefer this to the `for` as it lets us comp more transducers in easily.

```clojure
(defn get-md5s [config files]
  (if-let [output-dir (m/get-output-dir config)]
    (into {}
          (map (fn [f]
                 [f (-> (io/file output-dir f) (digest/md5))]))
          files)))
```

PR #172, opened 2019-03-11
Fix Leaver Prior

mattford63:
See the leaver prior notebook included for impact.


src/clj/witan/send/params.clj
```diff
@@ -5,7 +5,7 @@
             [witan.send.states :as s]))
 
 (defn joiner?
-  [[ay state-1 state-2]]
+  [[_ state-1 state-2]]
```
otfrom: leave as `ay` rather than as `_`

src/clj/witan/send/params.clj
```diff
@@ -80,11 +82,16 @@
                                    (assoc coll ay {:alpha (double (/ alpha total))
                                                    :beta (double (/ beta total))})))
                                {} observations)
-        prior-per-year (continue-for-latter-ays prior-per-year academic-years)]
+        unobserved-priors (reduce (fn [coll ay]
```
otfrom: `beta-params-leavers` is one of the functions where I struggle to know what each step of the let does. It would be great to factor out each step of the let into their own functions and have at least some simple tests on this to show what we expect each step to look like.

mattford63: Would love to, but pre-existing work depends on this at present.  Will gladly pick up after other fixes are merged.

PR #173, opened 2019-03-12
Mover Beta Prior Defaults

mattford63:
See impact in notebook here:

https://github.com/MastodonC/witan.send/blob/dc5adf28232741fc6c1e18df824b42663414cad6/data/demo/nbs/inference-movers-beta.ipynb

Acceptance tests won't pass until the results for checking dirs are updated.

The md5sums are consistent on validation results at 10, or a 1000 for the mover change.  The md5sums are not consistent for the distribution outputs as we'd expect.


mattford63:
We only do 10 runs, not a 100.  That can be bumped up.

Which sets of results do you want a percentage difference of?

Why do we care about this?  Say every result changes by 0.0001 - that's a 100% difference (in a naive implementation) between datasets.  Seems to me it's the charts that are important.

seb231:
So at 10 or 100 runs there can still be quite a bit of variation between runs, so to factor this out I suggest doing more (1000 is just the top end of what we usually do).

I'd like to see comparisons between all the average results if I'm honest, so each output dataset. This can then be rolled up and some summary stats (average, varaince, std.dev, maybe) calculated to summarise how different two datasets are from each other.

I agree the charts are a much better way of visualising the changes or lack thereof, but as Chris Owen has recently been pointing out, it can be quite easy to miss things in these charts. I'd like to see something like this: https://docs.google.com/spreadsheets/d/1YS_AvZtLITBCGzH9qiFOYaN9U9SWgeXMjGKVJyeKjWM/edit#gid=1923337926 (see cols delta-mean-ground-truth | delta-pc-of-ground-truth)

PR #174, opened 2019-03-14
Tdf/joiner unobserved ay

mattford63:
- See eng. blog post for discussion of issue. 
- See for an analysis https://github.com/MastodonC/witan.send/blob/8456d9ecabdda451526807f585d0f8ca160adf55/data/demo/nbs/inference-joiners-beta.ipynb
- Acceptance checks will fail until new golden results are added to the PR on approval


mattford63:
Can you check the code please to make sure you understand exactly what it means (as I also would like confirmation the logic is correct).

The notebook imports the metadata and config for the runs (see the section near the top)

src/clj/witan/send/params.clj
```diff
@@ -44,11 +44,16 @@
           params (sort academic-years)))
 
 (defn calculate-joiners-per-calendar-year
+  "The result maps keys are used to lookup up external population so need to be added
+   even though there is no observed count"
   [transitions]
+  (let [observed-calendar-years (into #{} (map :calendar-year transitions))
+        joiners-per-calendar-year (into {} (for [cy observed-calendar-years] [cy {}]))]
```
seb231: I'm not sure what this bit is doing?

src/clj/witan/send/params.clj
```diff
@@ -104,23 +107,24 @@
...
+       (reduce
+        (fn [coll cy]
+          (let [j (get-in joiners [cy ay])
+                p (get-in population [cy ay])]
+            (if j
+              (-> coll
+                  (update-in [ay :alpha] m/some+ (/ j n))
+                  (update-in [ay :beta] m/some+ (/ (- p j) n)))
+              (-> coll
+                  (update-in [ay :alpha] m/some+ 0.001) ; default prior may need tweaking
```
seb231: It looks like from your comment that you are unsure of this number, but I am presuming at this point it is just a nice low number as default?

mattford63: Yep.

src/clj/witan/send/params.clj
mattford63: It creates a set of observed calendar years from all the transitions data and then creates an empty set of results but with keys of each calendar year i.e, `{2016: {}, 2017: {}, 2018: {}}`

The results now look of the function now look something like `{2016: {1: 20, 2:34,..}, 2017: {}, 2018: {1: 20, 2:34,..}}` rather than `{2016: {1: 20, 2:34,..}, 2018: {1: 20, 2:34,..}}`.  Having this key present is enough for the later function that looks up the non-joiners for 2017 to work properly.

seb231:
Thanks for the confirmation @mattford63. Are you able to answer this one now: What does 86.16% better or equal to the current results actually mean?

PR #175, opened 2019-03-26
incorperate-needs

mattford63:
Tests are expected to fail until I update the "golden" version of the results to validate against.  Will do once reviewed (else it's too much change).

See here for analysis of impact

https://github.com/MastodonC/witan.send/blob/70eaf7d89c34eb091c3f4f2751cd0738c8e663f1/data/demo/nbs/inference-movers-dirichlet.ipynb

The code updates the mover-dirichlet calculations to use the valid settings and allows need movement.


seb231:
@mattford63 how come you are comparing 1000 sim run with 10?

mattford63:
> @mattford63 how come you are comparing 1000 sim run with 10?

An accident, now fixed.

src/clj/witan/send/params.clj
```diff
@@ -128,19 +117,20 @@
...
-                               (if (or (= setting-1 c/non-send)
+        observations (reduce (fn [coll {:keys [academic-year-1 need-1 need-2 setting-1 setting-2]}]
+                               (if (or (= setting-1 c/non-send) ; leaver or joiner
                                        (= setting-2 c/non-send))
                                  coll
-                                 (if (not= setting-1 setting-2)
-                                   (update-in coll [academic-year-1 (s/join-need-setting need-1 setting-1) :alpha] m/some+ 1)
-                                   (update-in coll [academic-year-1 (s/join-need-setting need-1 setting-1) :beta] m/some+ 1))))
+                                 (if (not (and (= setting-1 setting-2) ; not a remainer i.e a mover
+                                               (= need-1 need-2)))
```
seb231: Is this logic definitely correct? It's late in the day so sorry, but should it not be 
```(if (and (not (= setting-1 setting-2)) (not (= need-1 need-2))))```?

mattford63: To be a stayer you have to have setting-1 = setting-2 *and* have need-1 = need-2.  Everything else (it's negation) is a mover.

By DeMorgan's law your logic is not equivalent to this (the equivalent would be setting-1 != setting-2 *or* need-1 != need-2. i.e, at least one is different - which is what we want).

What you've suggested is that *both* have to be different which was the issue in the first place.

seb231: "To be a stayer you have to have setting-1 = setting-2 and have need-1 = need-2. Everything else (it's negation) is a mover." yep makes total sense in the light of day

mattford63:
If you look at the the heatmaps graphed in the analysis notebook (link in description above).  The interactive charts of the form `hm.mover_dirichlet_v2(results_dir)` you will see for some years the parameters aren't not calculated.  This is due to the valid states file prohibiting movement and so the new code does not actually generate them - however there is some uncertainty as to which year the parameters should be applied.  And in some cases the valid states file seems to be prohibiting a bit too much.

Investigate the heatmaps and codebase to decide when/where the parameters should be calculated.  Make the final failing dirichlet test pass based on the investigation above.

PR #176, opened 2019-03-29
Better historic filtering

seb231:


src/clj/witan/send/model/prepare.clj
```diff
@@ -189,6 +176,63 @@
...
+(defn build-pred [m]
+  (partial (:op (val m)) (:val (val m))))
+
+(def operators {:= =
+                :< <
+                :<= <=
+                :> >
+                :>= >=})
+
+(defn operator-key [filter-k]
```
otfrom: I don't think you need this function. I think you can use the operators map directly

src/clj/witan/send/model/prepare.clj
```diff
@@ -189,6 +176,63 @@
...
+
+   Each key much conform to :calendar-academic, :calendar-setting, :calendar-need, :academic-setting,
+   :academic-need or :setting-need.
+
+   The inner map must contain two key-value pairs, the key corresponding to an operator and the value
+   corresponding to a value to filter on"
+  (let [k (first filter-kv)
+        v (second filter-kv)
+        op1 (operator-key (first (keys v)))
+        op2 (operator-key (second (keys v)))
```
otfrom: these 2 uses of operator key could just be directly acting on the operators map

src/clj/witan/send/model/prepare.clj
```diff
@@ -189,6 +176,63 @@
...
+(def operators {:= =
+                :< <
+                :<= <=
+                :> >
+                :>= >=})
+
+(defn operator-key [filter-k]
+  (filter-k operators))
+
+(defn build-filter [filter-kv]
```
otfrom: I'd call this `remove-transitions-xf` as it returns a transducer function and is doing a remove rather than a filter. I'd also call `filter-kv` `remove-operation` as that is what it is defining and isn't *really* a kv at this point but a tuple.

src/clj/witan/send/model/prepare.clj
```diff
@@ -230,9 +274,7 @@
                               (transitions-map modified-transitions)
                               (transitions-map transitions))
          transitions-filtered (when filter-transitions-from
-                                (mapcat (fn [year] (filter #(> (:calendar-year %) year)
-                                                           (or modified-transitions transitions)))
-                                        [filter-transitions-from]))
+                                (reduce #(sequence (build-filter %2) %1) (or modified-transitions transitions) filter-transitions-from))
```
otfrom: that line is a bit of a mind bender. It is using a seq as a reducing function over `filtered-transitions-from` as input and using `(or modified-transitions transition)` as the seed state?

seb231: not quite, it's using `filtered-transitions-from` as input to build a reducing function to then apply to `(or modified-transitions transition)`.

PR #178, opened 2019-04-15
comparison plots with legends and zero-indexing

seb231:


seb231:
results are mislabelled in all outputs

PR #186, opened 2019-07-03
Show population lookup failure

otfrom:


src/clj/witan/send/params.clj
```diff
@@ -115,9 +115,12 @@
...
                   (update-in [ay :beta] m/some+ (/ (- p j) n)))
-              (-> coll
-                  (update-in [ay :alpha] m/some+ 0.001) ; default prior may need tweaking
-                  (update-in [ay :beta] m/some+ (/ p n))))))
+              (if p
+                (-> coll
+                    (update-in [ay :alpha] m/some+ 0.001) ; default prior may need tweaking
+                    (update-in [ay :beta] m/some+ (/ p n)))
+                (throw (ex-info "No population data for calendar year/academic year pair"
+                                {:calendar-year cy :academic-year ay}))))))
```
seb231: I was just looking to see if this or something similar should also be in other similar functions, but I see that other functions calculate `academic-years` and use it to bound which AYs there are params for. This function calculates the AYs but doesn't use them! Consider either refactoring to use them, or getting rid of line 106.

otfrom: I'm a bit confused as academic-years from the valid-states is used to drive the outer reduce here. We can't use transitions to generate it as there may be gaps in the transition data and we want to do a joiner calc for every ay/cy pair that *can* happen not just that we've observed in the transitions.

I'd like to add in a cross check for this in the validation so that we catch it in the data prep phase, but even now when the exception is thrown it is reasonably clear what is going on now.

seb231: AH! sorry I didn't expand the code below! this was the last thing I could see: `joiner-calendar-years))`. Apologies.

PR #194, opened 2019-10-02
Feature/goal seek

seb231:


src/clj/witan/send/goal_seek.clj
```diff
@@ -0,0 +1,61 @@
...
+   e.g. {:year 2019 :population [6 12]}"
+  (let [configs (map update-results-path (mc/generate-configs (create-transition-modifier-seq m start end step) base-config))]
+    (loop [n 0]
+      (let [config (nth configs n)
+            result (do (main/run-recorded-send config)
+                       (get-target-pop config))
+            diff (- (:population (first (filter #(= (:year target) (:year %)) result)))
+                    (:population (first (filter #(= (- (:year target) 1) (:year %)) result))))
+            _ (println diff)
+            _ (println n)]
```
otfrom: remove the `println`s if you are happy

src/clj/witan/send/goal_seek.clj
```diff
@@ -0,0 +1,61 @@
...
+            filepath))
+
+(defn get-target-pop [config]
+  (let [state (-> (get-in config [:transition-parameters :transitions-to-change])
+                  first
+                  (clojure.set/rename-keys {:setting-2 :setting :need-2 :need :academic-year-2 :ay})
+                  (select-keys [:setting :need]))
+        out-dir (:output-dir (:output-parameters config))
+        state-pop (str (System/getProperty "user.home")
+                       "/code/witan.send.afc/"
```
otfrom: not right for witan.send?

src/clj/witan/send/goal_seek.clj
```diff
@@ -0,0 +1,61 @@
...
+
+(defn update-results-path [config]
+  (assoc-in config [:output-parameters :output-dir] (str "results/" (get-in config [:output-parameters :output-dir]))))
+
+(defn target-results [m start end step base-config target]
+  "Takes a map of keys partially matching a transition, a start, end and step range to modify
+   the transition by, a template config and a map containing a target population range and year
+
+   e.g. {:year 2019 :population [6 12]}"
+  (let [configs (map update-results-path (mc/generate-configs (create-transition-modifier-seq m start end step) base-config))]
```
otfrom: reading from the side effecting and file writing proc bugs me, but that is where we are atm. 

I wonder about driving the loop off of the configs and doing 

```
(loop [configs (create-configs blah blah generate maps)]
  (let [[config & rest-configs] configs]
    (main/run-recorded-send config)
    (when (need-better-result? (get-target-pop config))
      (recur rest-configs))))
```

So I'd factor out the diff and predicate for your loop and a function that generates all your configs.

otfrom: and the way we should do it (eventually) is by taking the middle config, testing that, and if we need to go higher or lower passing the higher or lower configs to the recur.

src/clj/witan/send/goal_seek.clj
seb231: ah yes thank you

PR #196, opened 2019-10-10
More explicit and faster send run

otfrom:
The existing send run code was difficult to follow and debug, and the expensive reduce portion ran single threaded while the cheaper simulation bit was multi-threaded. 

This PR swaps this around. Get rid of the pmap, runs each reducing function in its own future and has explicit code for each data product produced in the reduce step.

There are some data shape dependencies based on how the old code used the reducing function composition that should be factored out in the future, but I think this is a simplification of what was there before with the advantage of being a bit faster to run as well.

witan.send.data-products replaces the roll up and calculate functions in witan.send.util `model-population-by-*` and the `reduce-rf` and `combine-rf` functions in witan.send.run. It makes it a bit easier to run each one in a future and possibly later by allowing us to create a data pipeline using core.async so that we can add more ways of re-purposing the simulation data.

### DIFF
--- a/src/clj/witan/send/model/data_products.clj
+++ b/src/clj/witan/send/model/data_products.clj
@@ -1,0 +1,193 @@
+(ns witan.send.model.data-products
+  (:require [witan.send.constants :as constants]
+            [witan.send.maths :as m]
+            [witan.send.schemas :as sc]
+            [witan.send.states :as states]
+            [witan.send.utils :as u])
+  (:import org.HdrHistogram.IntCountsHistogram))
+
+(def number-of-significant-digits 3)
+
+(defn roll-up-calendar-year-by-need-setting [calendar-year]
+  (transduce
+   (map (fn [[[ay need-setting] population]] [need-setting population]))
+   (fn
+     ([totals] totals)
+     ([totals [need-setting population]]
+      (update totals need-setting m/some+ population)))
+   {}
+   calendar-year))
+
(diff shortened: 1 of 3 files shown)


src/clj/witan/send/model/data_products.clj
```diff
@@ -0,0 +1,193 @@
...
+    (map (fn [[[ay need-setting] population]] [ay population]))
+    (map (fn [[ay population]] [(u/ay-groups ay) population])))
+   (fn
+     ([totals] totals)
+     ([totals [ay-group population]]
+      (update totals ay-group m/some+ population)))
+   {}
+   calendar-year))
+
+(defn histogram-summary-rf
```
seb231: This looks like code from utils, do you not need to also remove it from there?

src/clj/witan/send/model/data_products.clj
```diff
@@ -0,0 +1,193 @@
+(ns witan.send.model.data-products
```
seb231: why data-products?

src/clj/witan/send/model/data_products.clj
otfrom: I've removed the unused code from utils now

src/clj/witan/send/model/data_products.clj
otfrom: data products are the things we make from the main projection. I've not found a better name (happy to have one).

PR #199, opened 2019-10-24
Kondo fixes

otfrom:
I've used the clj-kondo linter to highlight a number of fixes. The fall into the following categories:

1. if without a second leg
2. misplaced docstrings
3. redundant lets
4. unused namespaces
5. :refer :all

Some unused vars have been left in where they are in function calls or destructuring as they serve as documentation for the structure. clj-kondo still reports these as warnings.


seb231:
let's party like it's PR 199!

PR #205, opened 2019-11-26
Allow modify transitions from or until a specific future year

seb231:
Previously `modify-transitions-from` only allowed the ability to modify transitions at a specific future calendar year and onward. Now a user can set when they'd like to start or end modifying transitions in a projection (note: not both)

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -2,7 +2,7 @@
 
 The config is an .edn file, essentially a map (`{}`) containing keys (`:foo`) and associated values. There are five sections to the config, defined as maps within the larger config map.
 
-As example config.edn can be found [here](https://github.com/MastodonC/witan.send/blob/master/data/demo/config.edn).
+As example config.edn can be found [here](../data/demo/config.edn).
 
 ### `:file-inputs`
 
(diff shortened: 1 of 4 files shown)


doc/config.md
```diff
@@ -51,9 +51,9 @@ Multiple entity pair filters can be applied in parallel.
...
+Expects a map with either `:from` or `:until` key and a calendar as an integer to modify transitions either from or until, when `:transitions-to-change` is used. Used for either a [“Modify setting(s) transitions rates”](https://github.com/MastodonC/witan.send/blob/master/doc/scenarios.md#modify-settings-transitions-rates), [“Modify setting(s) transitions rates and transfer individuals to alternative setting(s)”](https://github.com/MastodonC/witan.send/blob/master/doc/scenarios.md#modify-settings-transitions-rates-and-transfer-individuals-to-alternative-settings) or [“Modify transitions from a specific future calendar year”](https://github.com/MastodonC/witan.send/blob/master/doc/scenarios.md#modify-transitions-from-a-specific-future-calendar-year) scenario.
```
otfrom: relative links pls rather than github links

PR #210, opened 2020-05-28
create transition definitions

seb231:


doc/dsdr/0003-transition-definition.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Context
+
+We need an agreed upon definition of what a "transition" is. The model takes a ["transitions file"](data/demo/data/transitions.csv), which is a csv of all transitions each calendar year over two or more years.
+
+## Decision
+
+A transition is recorded in the transitions file composed of a calendar year, a starting academic year, a ending academic year, a need and a setting corresponding to the starting academic year and a need and setting corresponding to the ending academic year.
+
+A transition is a movement between calendar and academic years (these should always increase). Need and setting can change, but this is not prerequisite. A transition is typically built from two snapshots of where someone in SEND was in January of a particular year and then in the following January. The transition is labelled with the starting calendar year of the transitions, for example a transition between 2016 and 2017 would be labelled 2016.
```
otfrom, line 11: Insert: We have found in practice that the number of children staying back or skipping academic years hasn't been large enough, so we currently exclude them from the transitions file and modelling.

doc/dsdr/0004-joiner-definition.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Status
+
+In review.
+
+## Consequences
+
+The calendar year being moved to is assumed, not specified (in the example above, this would be 2017).
+
+Even though the "in SEND" information is on the "ending year" side (`need-2`, `setting-2`), a joiner transition is referred to be the calendar year stated in the transition data.
```
otfrom, line 21: ah, this is an important one for those of us creating charts. ;-)

doc/dsdr/0005-leaver-transition.md
```diff
@@ -0,0 +1,19 @@
+# Leaver
+
+## Context
+
+We need an agreed upon definition of what a "leaver" is. 
+
+## Decision
+
+A leaver is someone whose EHCP has ended or has moved out of the Local Authority and therefore the Local Authority are no longer responsible for their education and support. A leaver for 2016 would be someone who between Jan-16 and Jan-17 transitioned out of SEND.
```
otfrom, line 9: or has moved beyond the maximum year of the SEND system?

doc/dsdr/0008-need-transition.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Decision
+
+Need only changes are not included in the model. They are deemed uncommon and would increase the number of rates calculated, in short making the subpopulations smaller and harder to project.
+
+The witan.send model is designed to project a local authorities SEND population and allow them to model scenarios for the future to help them adjust planning. It is considered unlikely that modelling the rate at which people change needs would be a policy that a local authority could realistically implement.
+
+## Status
+
+Implemented.
```
otfrom, line 17: As a decision, I think this is proposed, but not part of the model, so I'd rather the status was Proposed (and we have a link to the PR above)

doc/dsdr/0005-leaver-transition.md
seb231: So I did have this but then removed it, as I think that for someone aging out they would qualify as "someone whose EHCP has ended". I purposefully didn't expand on why someone's EHCP might end, as LAs have different opinions on how it is/should be possible or not. Thoughts?

doc/dsdr/0008-need-transition.md
seb231: But the decision we made has been implemented/applied already?

doc/dsdr/0005-leaver-transition.md
otfrom: "whose EHCP has ended due to age or other reason"

PR #211, opened 2020-07-17
Handle exceptions in model around movers and leavers

otfrom:


src/clj/witan/send/model/run.clj
```diff
@@ -87,18 +87,28 @@
...
+      (apply-leavers-movers-for-cohort-unsafe population-by-state cohort params calendar-year
+                                              valid-transitions make-setting-invalid))
+    (catch Exception e
+      (throw (ex-info "Could not calculate leavers and movers for cohort."
+                      {:population-by-state population-by-state
+                       :cohort cohort
+                       :params params
+                       :calendar-year calendar-year
+                       :valid-transitions valid-transitions
+                       :make-setting-invalid make-setting-invalid}
```
seb231, line 110: why throw all this?

otfrom: it gives me a way to look at all of the inputs when it does throw an exception as I'm having difficulty debugging it otherwise. It is too big though and it would be good to find another way to handle it.

src/clj/witan/send/model/output.clj
```diff
@@ -162,23 +162,22 @@
...
-                              normalisation (normalisation-constant-dirichlet params)]
-                          (map (fn [p] (into []
-                                             (concat prefix
-                                                     (conj (mapv name (states/split-need-setting (key p)))
-                                                           (/ (val p) normalisation)
-                                                           normalisation
-                                                           (val p)))))
-                               params))))
-              [] alphas)
-      (sort)))
```
seb231, line 181: How come?

otfrom: it was getting a mix of string and number types in one of the fields (I didn't quite track it down). It doesn't seem to affect any results if you don't care about sorting as it writes it out to csv.

PR #212, opened 2020-08-03
Replace project.clj with deps.edn

seb231:
tests fails with clojure 1.10 so maintaining 1.9 for the moment


bin/koacha
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+clojure -J-Xmx8g -A:test -m kaocha.runner "$@"
```
otfrom, line 3: I think this file has the command we want, but the filename is wrong

deps.edn
```diff
@@ -0,0 +1,25 @@
+{:paths ["src/clj"]
+ :deps  {org.clojure/clojure  {:mvn/version "1.9.0"}
```
otfrom, line 2: add comment saying moving to 1.10 causes a failure

deps.edn
```diff
@@ -0,0 +1,25 @@
...
+ {:test {:jvm-opts ["-Xmx8g"]
+         :extra-paths ["test"]
+         :extra-deps  {org.clojure/test.check {:mvn/version "0.9.0"}
+                       lambdaisland/kaocha {:mvn/version "1.0.641"}}}
+  :runner
+  {:extra-deps {com.cognitect/test-runner
+                {:git/url "https://github.com/cognitect-labs/test-runner"
+                 :sha     "76568540e7f40268ad2b646110f237a60295fa3c"}}
+   :main-opts  ["-m" "cognitect.test-runner"
+                "-d" "test"]}}}
```
otfrom, line 25: Do we still need this bit?

PR #213, opened 2020-08-03
academic_years.clj

seb231:


src/clj/witan/send/domain/academic_years.clj
```diff
@@ -0,0 +1,68 @@
...
+              12 12
+              13 13
+              14 14
+              15 15
+              16 16
+              17 17
+              18 18
+              19 19
+              20 20
+              21 21
```
otfrom, line 30: I think we want to limit to AY 21

src/clj/witan/send/domain/academic_years.clj
```diff
@@ -0,0 +1,68 @@
...
+  (into (sorted-set) (inclusive-range 7 9)))
+
+(def key-stage-4
+  (into (sorted-set) (inclusive-range 10 11)))
+
+(def key-stage-5
+  (into (sorted-set) (inclusive-range 12 14)))
+
+(def ncy-15+
+  (into (sorted-set) (inclusive-range 15 25)))
```
otfrom, line 58: limit this to 20 and create an `outside-of-send-age` set that has 21 in it?

PR #215, opened 2020-09-17
Capture the model state for inspection w/o blowing up emacs.

otfrom:
You can now look at the exception state here by doing

```clojure
(require 'witan.send.model.run)

(deref witan.send.model.run/exception-info)
```

and doing M-x cider-inspect-last-sexp on the deref


seb231 (approved):
LGTM

PR #216, opened 2020-10-03
Output simulated transitions

otfrom:
The basis of the send model is creating multiple simulations where each simulation is a count of transitions. This change outputs those simulated transitions for future analysis.


seb231 (approved):
That seemed too easy... 🤔

otfrom:
It was just there waiting for us. I got lucky that I found it.

PR #218, opened 2020-10-14
The model generates the *-2 side for ay and cy

otfrom:
So we need to fix it when creating the *-1 side for ay and cy.


src/clj/witan/send/model/output.clj
```diff
@@ -106,13 +106,16 @@
                                                     (m/round0 (/ transitions simulations))]))))))]
       (csv/write-csv writer (into [headers] rows)))))
 
-
-(defn format-transition-count [idx [[cy ay need-setting-1 need-setting-2] transition-count]]
+;; TODO: Output this as a vec of maps as nippy
+(defn format-transition-count
+  "Each transtion count comes throug as the *-2 side for things like cy
```
seb231, line 111: throug = through

seb231 (approved):
👍

src/clj/witan/send/model/output.clj
otfrom: fixed

PR #219, opened 2020-10-15
Log the config on an exception in run-recorded-send

otfrom:
So we can make sure any config transformations are successful.


seb231 (approved):
:shipit:

PR #220, opened 2020-10-21
Fix/better goal seek options

seb231:


src/clj/witan/send/goal_seek.clj
```diff
@@ -25,35 +31,40 @@
...
-                       out-dir
-                       "/Output_State_pop_only.csv")]
-    [state state-pop]))
+(defn match-need-setting [data state]
+  (= (second (key data)) (st/join-need-setting (:need state) (:setting state))))
+
+(defn match-ay [data state]
+  (let [ay-state (:academic-year state)]
+    (if (coll? ay-state)
+      (true? (some (fn [ay] (= (first (key data)) ay)) ay-state))
```
otfrom, line 40: Does this need `true?` or will truthy do?

src/clj/witan/send/goal_seek.clj
```diff
@@ -25,35 +31,40 @@
...
+                        (every? (fn [k] (contains? state k)) [:setting :need])
+                        (match-need-setting % state)
 
-(defn update-results-path [config]
-  (assoc-in config [:output-parameters :output-dir] (str "results/" (get-in config [:output-parameters :output-dir]))))
+                        (contains? state :academic-year)
+                        (match-ay % state)))
+         (map val)
+         (apply merge-with +)
+         :median)))
```
otfrom, line 64: I presume at some point this will need to be a real median?

src/clj/witan/send/goal_seek.clj
```diff
@@ -80,59 +94,133 @@
...
+              (recur new-modifier
+                     (into tested-modifiers [new-modifier])
+                     (into population [pop-result])))
+
+            (within-pop-range? target-pop-range pop-result)
+            (let [config (update-results-path config results-path)]
+              (so/output-send-results projection (:output-parameters config))
+              (main/save-runtime-config config)
+              (main/save-runtime-metadata config (md/metadata config))
+              )))))))
```
otfrom, line 222: This will do for now, but feels a bit brittle. We should look into this in future versions.

src/clj/witan/send/main.clj
```diff
@@ -79,4 +79,8 @@
   ([config-path]
    (-> config-path
        read-config
-       run-recorded-send)))
+       (run-recorded-send true)))
```
otfrom, line 82: can't we just add `:print-warnings?` to the config and assoc it onto config here if we want rather than adding an arity with a default of true?

otfrom (approved):
A few things to tidy up if you want

src/clj/witan/send/goal_seek.clj
seb231: Had to debug and check, but yes because we want at least one (hence `some`) of `(first (key data))` to equal `ay` this is actually testing for `true?`

src/clj/witan/send/goal_seek.clj
seb231: ah are you pointing out that I'm summing medians...

src/clj/witan/send/goal_seek.clj
seb231: What are your issues? Too big? `cond` to complicated?

src/clj/witan/send/main.clj
seb231: The default is currently true. I found this provided me with the most flexibility

src/clj/witan/send/goal_seek.clj
otfrom: streaming out the csv like this would mean we might lose it all if there was an exception or other error

src/clj/witan/send/goal_seek.clj
otfrom: yes

seb231: yes, then, although for this purpose, providing the previos projection has been run 1000 times, this is sufficient, as it's just used as a check and basic to start the modifier from

src/clj/witan/send/goal_seek.clj
seb231: ah okay, so interestingly I wonder if this is actually causing an issue. My run last night found the best value and output the results but then the process never ended, I wonder if it was because the stream was still open. Do you have any immediate suggestions for how to improve this?

src/clj/witan/send/main.clj
otfrom: OK

PR #221, opened 2020-11-19
Improve the wall clock performance of the model by using more cores.

otfrom:


src/clj/witan/send/model/data_products.clj
```diff
@@ -145,39 +145,54 @@
...
+                                  simulations)))
+        total-in-send-by-need (future (summarise-results
+                                       "Total in SEND by Need"
+                                       roll-up-calendar-year-by-total-in-send-by-need simulations))
+        population-by-setting (future (summarise-results
+                                       "Total in SEND by Setting"
+                                       setting-summary/population simulations))
         cost-by-setting (future (into []
-                                      (summarise-results (partial setting-summary/cost cost-lookup) simulations)))]
-    {:by-state @by-state
```
seb231, line 173: Why do we not do by-state any more? This is causing this test to fail: expected-validation-results. So it either needs fixing or by-state needs adding back in

seb231 (changes requested):
TLDR: Tests fail, but are also not good enough. Need fixing and improving.

Tests currently fail.

1. There is a test on output_state.csv which is no longer produced, needs removing.
2. There is a test on output files which are being tested against before they are output
3. There is  a checksum test on output files which says there is a change but it is not obvious where it is. This is cause by the change from clj1.9 to 1.10. Suspect it has something to do with random number generation. Requires tests

We agreed that we need new tests, similar to the above failing tests, which instead test the results of a projection in memory rather than output files.

Within this scope also should be removal of the current output in favour of just returning the data inputs, simulations and training information (params)

src/clj/witan/send/model/output.clj
```diff
@@ -258,7 +272,19 @@
...
            keep-temp-files? use-confidence-bound-or-interval population-file]}]
-  (let [transform-transitions (->> transitions
+  (println "Outputting Results")
+
+  ;; Output the rest of the results
+  (let [dir (str project-dir "/" output-dir)
+        _ (when (not (.isDirectory (io/file dir)))
+            (.mkdir (io/file dir)))
+        ;; Output the big simulation results
+        _ (a/thread (output-simulated-transitions (str project-dir "/" output-dir) "simulated-transitions" @simulated-transitions))
```
seb231, line 282: I'm not sure where there is a `thread` here for the whole process? I think I understand the `thread` within `output-simulated-transitions`

seb231: Oh I think it's because there are more threads downstream

PR #224, opened 2021-01-19
Update testing

seb231:
Deleted the redundant tests, pulled across the tests that just “worked” from the [performance branch](https://github.com/MastodonC/witan.send/pull/221), made sure kixi.stats, hdrhistogram and clojure are the old versions (as per master), and added the distributions and run model tests, which failed with the old deps, and have “fixed” them, so that the tests pass with our current deps.

Would be worth checking I've brought all of the test specific code over from the performance branch. 

Test still fail with Clojure 1.10 as expected.


otfrom (approved):
Thx. Good harvesting and a good place to stop. Good narrative in the commits as well.

PR #226, opened 2021-05-19
Enhanced AY domain logic

otfrom:
I needed to be able to do primary/secondary and outside of send ages


seb231:
Looks good 👍

I’d not heard of “First School”? You could also have Sixth Form, which would be NCY12-13

PR #228, opened 2021-09-16
Added a way to classify various year groups

otfrom:


seb231 (approved):
Happy with this, though I do feel like theirs probably a better name for the fn but every suggestion I had could also be confusing

PR #230, opened 2022-01-27
More ncy predicates and domain helpers

otfrom:


seb231 (approved):
Annoyingly I've just been making changes to the ns too

PR #232, opened 2022-02-01
Goal seek update

seb231:


deps.edn
```diff
@@ -15,7 +15,8 @@
          org.apache.commons/commons-math3 {:mvn/version "3.6.1"}
          ;; upgrading to 2.1.10 causes a test to fail
          org.hdrhistogram/HdrHistogram    {:mvn/version "2.1.9" #_"2.1.12"}
-         same/ish                         {:mvn/version "0.1.4"}}
+         same/ish                         {:mvn/version "0.1.4"}
+         witan.send.adroddiad/witan.send.adroddiad {:local/root "../witan.send.adroddiad"}}
```
otfrom, line 19: yeah, I'm worried about this dependency. I think we'll get away with it as we use adroddiad as a library, but you are right that we should split this out of the model.

src/clj/witan/send/domain/academic_years.clj
```diff
@@ -93,6 +93,9 @@
 (def secondary-school
   (into key-stage-3 key-stage-4))
 
+;; Key Stages 1 to 5
+(def school-age (reduce #(into %1 %2) (sorted-set) [primary-school secondary-school key-stage-5]))
```
otfrom, line 97: this is handy. thx

src/clj/witan/send/goal_seek.clj
```diff
@@ -9,7 +9,11 @@
             [witan.send.model.output :as so]
             [witan.send.metadata :as md]
             [clojure.java.io :as io]
-            [clojure.data.csv :as csv]))
+            [clojure.data.csv :as csv]
+            [tablecloth.api :as tc]
+            [witan.send.adroddiad.simulated-transition-counts :as stc]
+            [witan.send.adroddiad.simulated-transition-counts.io :as stcio]
+            [witan.send.adroddiad.summary :as summary]))
```
otfrom, line 16: this would be fine if goal-seek was a separate lib

src/clj/witan/send/goal_seek.clj
```diff
@@ -98,14 +102,13 @@
...
-                     (every? identity (witan.send.model.prepare/test-predicates % state-map))))
-       (map #(select-keys % [:population]))
-       (apply merge-with +)
-       :population))
+(defn get-baseline-population [baseline year state-map]
+  (let [result (->> baseline
+                    (filter #(and (= (:calendar-year %) year)
+                                  (every? identity (witan.send.model.prepare/test-predicates % state-map))))
+                    (map #(select-keys % [:population :median]))
+                    (apply merge-with +))]
```
otfrom, line 110: I think the data is small enough at this point that it doesn't matter that it is ->> rather than a transduce.

src/clj/witan/send/goal_seek.clj
```diff
@@ -98,14 +102,13 @@
...
+    (get result :population (get result :median))))
```
otfrom, line 111: Do we need to use `summary` if we only want the median for `goal-seek`?

src/clj/witan/send/goal_seek.clj
```diff
@@ -98,14 +102,13 @@
...
-       csv->state-pop
-       (filter #(and (= (:calendar-year %) year)
-                     (every? identity (witan.send.model.prepare/test-predicates % state-map))))
-       (map #(select-keys % [:population]))
-       (apply merge-with +)
-       :population))
+(defn get-baseline-population [baseline year state-map]
+  (let [result (->> baseline
+                    (filter #(and (= (:calendar-year %) year)
+                                  (every? identity (witan.send.model.prepare/test-predicates % state-map))))
```
otfrom, line 108: This feels like the important line I need to understand and I don't. I'm not sure why you are selecting things here to be summed later.

src/clj/witan/send/goal_seek.clj
```diff
@@ -148,19 +151,18 @@
...
         projection (s/run-send-workflow config false)
-        base-year (+ 1 (apply max (into (sorted-set) (map :calendar-year (:transitions projection)))))
-        output (:send-output projection)
-        result (into {} (map #(assoc {} %1 %2) (range base-year (+ base-year (count output))) output))
-        target-year-result (get-in result [target-year :by-state])
-        target-pop-result (get-target-pop config target-year-result)]
+        result (tc/dataset (eduction stcio/simulated-transitions->transition-counts-xf (:simulated-transitions projection)))
+        census (stc/transition-counts->census-counts result (apply min (:calendar-year result)))
+        summary (summary/seven-number-summary census [:calendar-year :setting :need :academic-year] :transition-count)
+        target-pop-result (get-baseline-population (tc/rows summary :as-maps) target-year (dissoc m :modify-transition-by))]
```
otfrom, line 157: I think it is possible to write `get-baseline-population` as a tmd/tc thing that might run faster.

src/clj/witan/send/model/prepare.clj
```diff
@@ -224,13 +224,17 @@
...
 
 (defn filter-transitions [pred-map transitions ages year]
-  (let [result (filter (fn [t] (every? identity (test-predicates t pred-map))) transitions)]
-    (if result
-      result
+  (let [pred-map (-> pred-map
+                     (rename-keys {:setting :setting-2, :need :need-2 :academic-year :academic-year-2})
+                     (update :setting-2 keyword)
+                     (update :need-2 keyword))
+        result (filter (fn [t] (every? identity (test-predicates t pred-map))) transitions)]
```
otfrom, line 231: I'm not sure I understand why this is happening.

otfrom (changes requested):
I think I understand most of what is going on here. I think this probably belongs in another library and there are a few changes I'd like to make to it so we aren't doing things like writing the data out and re-reading it.

I wonder if the first step would be to make it so that `run-send-model` returns better things that we can report on in better ways using adroddiad, but also allow us to access what is there in memory if we want to. (which would dump almost all of witan.send.output)

src/clj/witan/send/domain/academic_years.clj
otfrom: @seb231 I'd like to merge this bit even if we don't merge the rest but move it to another library

seb231:
I think we already discussed this, but goal-seek doesn't output any files (or read in again) until it's found the optimum modifier

seb231:
Agreed that this needs to not be merged and moved to it's only library

src/clj/witan/send/goal_seek.clj
seb231: No probably not. Didn't really think about the fact that we could calculate it separately.

PR #233, opened 2022-02-08
Deconstructed send

otfrom:
# Problem Statement

The current construction of the SEND model does lots of data manipulation inside particular functions in the overall run pipeline. This makes it very difficult to figure out what is happening in each step and inspect the results, and makes it difficult to change the flow, by either directly manipulating things or by changing some of the inputs.

Furthermore, many outputs are calculated that aren't used by our current reporting system. It is also difficult to access the simulations in RAM instead we tend to re-read them from disk.

# Proposal

Tease apart loading the inputs, training the model, and running the model. While the first attempt may not be flexible enough, it should expose enough of the workings that we become comfortable with changing the transitions we learn from, changing the learned distributions from the transitions and working more quickly with the results in tools like goal-seek.

# Goal

A more flexible model that is faster to run allowing us to do more exploration around results and scenarios.

### DIFF
--- a/src/clj/witan/send.clj
+++ b/src/clj/witan/send.clj
@@ -89,43 +89,46 @@
                                      (r/projected-future-pop-by-year projected-population seed-year))])
        (range simulations)))
 
+
+(defn unpack-transition-counts [{:keys [transitions simulation]}]
+  (map (fn [map-entry]
+         (let [key (first map-entry)
+               transition-count (second map-entry)
+               [need-1 setting-1] (-> key
+                                      (nth 2)
+                                      (states/split-need-setting))
+               [need-2 setting-2] (-> key
+                                      (nth 3)
+                                      (states/split-need-setting))]
+           {:calendar-year (dec (nth key 0))
+            :calendar-year-2 (nth key 0)
+            :academic-year-1 (dec (nth key 1))
+            :academic-year-2 (nth key 1)
+            :need-1 need-1
+            :setting-1 setting-1
+            :need-2 need-2
+            :setting-2 setting-2
+            :transition-count transition-count
+            :simulation simulation}))
+       transitions))
+
+(defn add-simulation-idx [simulation-idx sd]
+  (assoc sd :simulation simulation-idx))
+
 (def ->dataset-seq-xf
   (comp
    (map (fn [[simulation-idx simulation-data]]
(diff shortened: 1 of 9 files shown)


src/clj/witan/send.clj
```diff
@@ -0,0 +1,193 @@
...
+           (* 3 (quot (.availableProcessors (Runtime/getRuntime)) 4))
+           to
+           ->dataset-xf
+           (async/to-chan!! simulations))
+          (async/<!! (async/into [] to)))
+        (tc/dataset)
+        (tc/convert-types {:academic-year :int16
+                           :calendar-year :int16
+                           :simulation :int16
+                           :transition-count :int32})))
```
seb231: How come these are still in here?

src/clj/witan/send.clj
```diff
@@ -0,0 +1,193 @@
...
+                        :valid-states (wivs/csv->valid-states (str project-dir "/" valid-states))}]
+    (if settings-to-change
+      (assoc input-datasets :settings-to-change (wistc/csv->settings-to-change (str project-dir "/" settings-to-change))) ;; do we need this?
+      input-datasets)))
+
+(defn train-model [{:keys [input-datasets config print-warnings?]
+                    :as state}]
+  (p/prepare-send-inputs input-datasets (:transition-parameters config) print-warnings?))
+
+#_(defn run-model [{:keys [model config]
```
seb231: should this be removed now?

seb231 (approved):
I think my comments may be on old diffs as I was looking through commit by commit, so you may just be able to ignore them 🤷 

I think it looks good, tidier, and I assume more concise. 

This makes the processing more straight forward but doesn't change any output right? Does it also speed things up now?

src/clj/witan/send.clj
otfrom: that should be gone now

src/clj/witan/send.clj
otfrom: That should be gone now as well

otfrom:
Much faster, mostly by doing less work (such as no output other than the parquet). Much more configurable. The results will be different because the random numbers are side effecting, but it still produces roughly the same output.

I'd like to stabilise the random numbers in another PR, but I need to understand how they work a bit better, or at least refresh my memory.

PR #236, opened 2022-02-16
Define school ages

seb231:


src/clj/witan/send/domain/academic_years.clj
```diff
@@ -93,6 +93,10 @@
 (def secondary-school
   (into key-stage-3 key-stage-4))
 
+;; Key Stages 1 to 5
+(def school-age
+  (reduce #(into %1 %2) (sorted-set) [primary-school secondary-school key-stage-5]))
```
otfrom, line 98: This works. I'd probably write this like

```
(into (sorted-set) cat [primary-school secondary-school key-stage-5])
```

otfrom (approved):
slight style difference, but your way is fine.

PR #241, opened 2022-06-09
Check for existing resources

seb231:


src/clj/witan/send/parquet.clj
```diff
@@ -94,6 +97,10 @@
   saturate the cores of whatever machine is running the simulation.
   "
   [{:keys [config] :as projection-config}]
+  (when (false? (.exists (io/file "resources")))
+    (.mkdir (java.io.File. "resources")))
+  (io/copy (io/file (io/resource "logback.xml"))
+           (io/file "resources/logback.xml"))
```
otfrom, line 103: I don't think this will work as the logback.xml won't get loaded from resources. I'll have to check that though. We could have it throw and exception with a message to say that you should put in the file and re-run everything.

otfrom (changes requested):
I'd like to see this fix working wrt the logback.xml

src/clj/witan/send/parquet.clj
seb231: ah I understand your concern now with your fix!

seb231 (commented):
I can't approve as I'm a contributor, but I do

PR #242, opened 2022-07-15
Add out-dir to message.

eraderna:


otfrom (approved):
hooray for lsp-consult finding things.

PR #245, opened 2022-08-16
Add school-phase

eraderna:
Add defs and defns to return school-phase keywords and names given NCY, using nursery terminology for NCY -5 to -1, plus supporting defs for post-16 and post-19.

### DIFF
--- a/src/clj/witan/send/domain/academic_years.clj
+++ b/src/clj/witan/send/domain/academic_years.clj
@@ -97,26 +97,61 @@
 (def nursery
   (disj early-years 0))
 
-;; Primary school - Reception + Years 1 to 6
+;; Primary school - Reception and Key Stages 1 + 2 - Reception + Years 1 to 6
 (def primary-school
   (into (sorted-set 0) (into key-stage-1 key-stage-2)))
 
-;; Infant school - Reception + Years 1 and 2
+;; Infant school - Reception and Key Stage 1 - Reception + Years 1 to 2
 (def infant-school (into (sorted-set) (inclusive-range 0 2)))
 
-;; Junior school - Years 3 to 6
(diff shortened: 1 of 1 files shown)


src/clj/witan/send/domain/academic_years.clj
```diff
@@ -97,26 +97,61 @@
...
+   :primary   "Primary"
+   :secondary "Secondary"
+   :post-16   "Post 16"
+   :post-19   "Post 19"})
+
+(defn ncy->school-phase-name
+  "National Curriculum Year to School Phase name"
+  [ncy]
+  (school-phase-names (ncy->school-phase ncy))
+  )
```
otfrom, line 151: This is interesting and I'm probably being picky but I have a few things that make me itchy about this:

1. tidy up the trailing paren
2. I'd probably do this as `(-> ncy ncy->school-phase school-phase-names)`
3. Given that it is just a thread first getting the sweet names from the map I'm not sure I would have done it unless I knew I was going to use it a lot. I'm wondering if we should do though to prevent us from making the same function in our application code.

otfrom (approved):
Tidying up the trailing paren would be nice, but otherwise I think this is good to merge despite my comments and thoughts.

src/clj/witan/send/domain/academic_years.clj
eraderna: Thanks: Have updated to tidy up trailing paren and change to thread first rather than nest functions. Agree this is a relatively minor function but assuming we're happy with the school-phase-names then I do think we be needing to map from ncy through to names sufficiently often to warrant centralising in this library.

otfrom: :cool: :shipit:

PR #246, opened 2022-11-24
Removing age defs from domain.academic-years

eraderna:
- Delete ncy->age and age->ncy which were using an offset of 5 (i.e. age at end of school year, or average age mid-school-year) rather than 4. For mose DfE uses age on 31-Aug prior to the start of the school year is relevant: for that use new witan.send.adriddiad.ncy/ncy->age-at-start-of-school-year.
- Delete age-group as these are defined in witan.send.adroddiad.dsg-plan.


otfrom (approved):
:+1:

PR #247, opened 2023-08-21
Throw error if joiners or population data missing

seb231:
This will throw the same error message which is helpful, when either joiner and/or population is not present


otfrom (approved):
:+1:

